PR #3, opened 2022-11-23
Use lower case filename in main.yml

danh-arm:
The changes in #1 caused the pdf archiving in the CI to fail. Use lower case output filename to match the changes in #1.


jmarinho (approved):
Looks good to me.

sjg20 (approved):

PR #4, opened 2022-11-30
Spec suggestions as discussed on TF-A mailing list

jwerner-chromium:

PR #6, opened 2023-02-22
Document algorithm guidelines for standard operations

jwerner-chromium:
This patch adds a section to the specification that describes the steps needed to perform all basic operations on the TL. This is intended to provide further clarity by example on the rest of the specification and offer a guideline for implementers to follow.


source/transfer_list.rst
```diff
@@ -146,6 +146,221 @@ The TE header is defined in :numref:`tab_te_header`.
...
+
+#. Check that `available_size` is larger than 16, otherwise abort.
+
+#. Set `tl.signature` (`tl_base_addr + 0x0`) to `0x6e_d0ff`.
+
+#. Set `tl.checksum` (`tl_base_addr + 0x4`) to 0 (for now).
+
+#. Set `tl.version` (`tl_base_addr + 0x5`) to 1.
+
+#. Set `tl.hdr_size` (`tl_base_addr + 0x6`) to 16.
```
sjg20, line 322: Should we use hex consistently? E.g. 0x10 here?

source/transfer_list.rst
```diff
@@ -146,6 +146,221 @@ The TE header is defined in :numref:`tab_te_header`.
...
+
+- `tl_base_addr`: Base address of the existing TL.
+- `target_base`: Base address of the target region to relocate into.
+- `target_size`: Total size in bytes of the target region to relocate into.
+
+#. Calculate `alignment_mask` as `(1 << tl.alignment) - 1`.
+
+#. Calculate the current `alignment_offset` as `tl_base_addr & alignment_mask`.
+
+#. Calculate `new_tl_base` as `(target_base & ~alignment_mask) + alignment_offset`.
```
sjg20, line 346: Would this work?

new_tl_base = ((target_base + alignment_mask) & ~alignment_mask) + alignment_offset`.

sjg20 (approved):
A few thoughts but looks good

source/transfer_list.rst
```diff
@@ -146,6 +146,221 @@ The TE header is defined in :numref:`tab_te_header`.
...
+#. While `te_base_addr - tl_base_addr` is smaller or equal to `tl.size`:
+
+   #. *(optional)* Check that `te_base_addr + te.hdr_size + te.data_size - tl_base_addr`
+      is smaller or equal to `tl.size`, otherwise abort (the TL is corrupted).
+
+   #. If `te.tag_id` (`te_base_addr + 0x0`) is a known tag, interpret the data
+      at `te_base_addr + te.hdr_size` accordingly. (Do not hardcode the value
+      for `te.hdr_size`!)
+
+   #. Add `te.hdr_size + te.data_size` to `te_base_addr`.
```
jmarinho, line 202: Is this line meant to state how to compute the next TE address?
Should it say something like: "The next TE is at align8(te.hdr_size + te.data_size + te_base_addr)"
?

source/transfer_list.rst
```diff
@@ -146,6 +146,221 @@ The TE header is defined in :numref:`tab_te_header`.
...
+Inputs:
+
+- `tl_base_addr`: Base address of the TL to add a TE to.
+- `new_tag_id`: ID number of the tag for the new TE.
+- `new_data_size`: Size in bytes of the data to be encapsulated in the TE.
+- [data]: Data to be copied into the TE or generated on the fly.
+
+#. *(optional)* Follow the steps in `Reading a TL`_ to look for a TE where
+   `te.tag_id` is 0 (XFERLIST_VOID) and `te.data_size` is greater or equal to
+   `new_data_size`. If found:
```
jmarinho, line 216: Should we accomodate for the alignment requirements of the new TE?
For instance, if this new TE requires new_alignment, then we can place it in a XFERLIST_VOID iff "te.data_size >= new_data_size + padding", where padding = te_base_address mod new_alignment.

source/transfer_list.rst
```diff
@@ -146,6 +146,221 @@ The TE header is defined in :numref:`tab_te_header`.
...
+
+Inputs:
+
+- `tl_base_addr`: Base address of the existing TL.
+- `target_base`: Base address of the target region to relocate into.
+- `target_size`: Total size in bytes of the target region to relocate into.
+
+#. Calculate `alignment_mask` as `(1 << tl.alignment) - 1`.
+
+#. Calculate the current `alignment_offset` as `tl_base_addr & alignment_mask`.
```
jmarinho, line 344: should this be rather "alignment_offset = target_base & alignment_mask" ?

source/transfer_list.rst
jmarinho: I may be wrong, but should this be simply;
"new_tl_base = ((target_base + alignment_mask) & ~alignment_mask)" ?

jmarinho: example:
 target_base = 0x1008
 tl.alignment = 0x20 -> alignment_mask = 0x1f
 ~alignment_mask = 0xffe0

 new_tl_base = (0x1008 + 0x1f) & 0xffe0 = 0x1027 & 0xffe0 = 0x1020
 
 we probably do not need the "alignment _offset" variable and the line 348 below.

source/transfer_list.rst
```diff
@@ -146,6 +146,221 @@ The TE header is defined in :numref:`tab_te_header`.
...
+- `tl_base_addr`: Base address of the TL to add a TE to.
+- `new_tag_id`: ID number of the tag for the new TE.
+- `new_alignment`: The alignment boundary as a power of 2 that the data must be aligned to.
+- `new_data_size`: Size in bytes of the data to be encapsulated in the TE.
+- [data]: Data to be copied into the TE or generated on the fly.
+
+#. Calculate `alignment_mask` as `(1 << new_alignment) - 1`.
+
+#. If `(tl_base_addr + tl.size) & alignment_mask` is not 0, follow the steps in
+   `Adding a new TE`_ with the following different inputs (bypass the option to
```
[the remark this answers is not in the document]
jwerner-chromium: Done

source/transfer_list.rst
jwerner-chromium: Done

source/transfer_list.rst
jwerner-chromium: Agreed, this is simpler, thanks!

source/transfer_list.rst
jwerner-chromium: The whole thing is supposed to be a while loop in pseudocode. The indented part is the inner part of the loop. Do you think that's not clear enough (would it help to put a `#. Repeat` at the end or something like that)?

source/transfer_list.rst
jwerner-chromium: No longer needed.

source/transfer_list.rst
jmarinho: I understood it now, we're accumulating in te_base_addr.
it's effectively te_base_addr = te_base_addr + te.hdr_size + te.data_size.

Looks good to me!

jmarinho:
Looks good to me

jmarinho (dismissed):

sjg20 (approved):

danh-arm (dismissed):
Just minor comments

source/transfer_list.rst
```diff
@@ -146,6 +146,219 @@ The TE header is defined in :numref:`tab_te_header`.
...
+
+Inputs:
+
+- `tl_base_addr`: Base address of the existing TL.
+
+#. Compare `tl.signature` (`tl_base_addr + 0x0`) to `0x6e_d0ff`. On a mismatch,
+   abort (this is not a valid TL).
+
+#. Compare `tl.version` (`tl_base_addr + 0x5`) to the expected version
+   (currently `0x1`). On a `0x0`, abort (this is not a valid TL). On any other
```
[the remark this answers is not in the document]
jwerner-chromium: I factored all occurrences of the current version number out into a substitution reference to avoid the maintenance hazard.

source/transfer_list.rst
```diff
@@ -146,6 +146,219 @@ The TE header is defined in :numref:`tab_te_header`.
...
+   hardcode the value for `tl.hdr_size`!)
+
+#. While `te_base_addr - tl_base_addr` is smaller or equal to `tl.size`:
+
+   #. *(optional)* Check that `te_base_addr + te.hdr_size + te.data_size - tl_base_addr`
+      is smaller or equal to `tl.size`, otherwise abort (the TL is corrupted).
+
+   #. If `te.tag_id` (`te_base_addr + 0x0`) is a known tag, interpret the data
+      at `te_base_addr + te.hdr_size` accordingly. (Do not hardcode the value
+      for `te.hdr_size`!)
```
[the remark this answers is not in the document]
jwerner-chromium: Done

danh-arm (approved):

jmarinho (approved):

source/transfer_list.rst
```diff
@@ -146,6 +146,224 @@ The TE header is defined in :numref:`tab_te_header`.
...
+
+- `tl_base_addr`: Base address of the existing TL.
+
+#. Compare `tl.signature` (`tl_base_addr + 0x0`) to `0x6e_d0ff`. On a mismatch,
+   abort (this is not a valid TL).
+
+#. Compare `tl.version` (`tl_base_addr + 0x5`) to the expected version
+   (currently |current_version|). If there is an exact match, the TL is valid
+   for all operations outlined in this section. If `tl.version` is larger, the
+   TL is valid for reading but must not be modified or relocated. If
```
danh-arm, line 173: Sorry, I know this is now merged but I'd just like to question this statement. If we can assume that all future versions will be backwards compatible, then why is not safe to modify or relocate the TL according to this version of the spec? Not being able to at least relocate a TL could be problematic, for example when passing a TL through a component that has not upgraded to the latest version but needs to make a deep copy (e.g. due to the original memory location being inaccessible to a subsequent component).

sjg20: It is true that this could be quite a large limitation, but increasing the version may introduce a new restriction or feature that the old software cannot support.

The implication here is that the software needs to be updated. I can certainly see a problem here though, since it means that an update of an earlier component can break later phases.

jwerner-chromium: I... huh. Strange. I thought we had agreed that it would be backwards-compatible only for reading but not for writing (I don't think it was my idea), but I can't find any source for that anymore. Maybe I just totally misread your comment here (which actually says the opposite): https://github.com/FirmwareHandoff/firmware_handoff/pull/4#discussion_r1040852918

This is a trade-off that can go either way: if we say that older implementations are still allowed to modify new TLs, then that limits the amount of things we can add to the spec as a backwards-compatible update later. For example, if for whatever reason we one day decide that in addition to the simple checksum the TL header should now contain a SHA256 of the whole TL, that would be fine as a backwards-compatible update if older implementations have to treat the TL as read-only, but it would break if they're also allowed to append (because then they wouldn't know to update that hash). Of course, on the other hand not being able to append limits the amount of useful interop you can have with older implementations.

I don't have a strong opinion on this either way. If you say you prefer that older implementations be allowed to write as well, I'm happy to write another PR to change that. We could even split the version number into 4 and 4 bits if we want to potentially allow both (although that starts getting a bit complicated).

Allowing only relocation and reading is also possible, that just means you cannot make any changes in newer versions that would potentially affect relocation behavior (e.g. if we wanted to add a "relocation_count" field that counts how often the TL was relocated for some reason). It's probably pretty unlikely that that would ever come up but it's hard to predict future requirements perfectly.

source/transfer_list.rst
```diff
@@ -146,6 +146,224 @@ The TE header is defined in :numref:`tab_te_header`.
...
+   `tl.version` is smaller, either abort or switch to code designed to
+   interpret the respective previous version of this specification (note that
```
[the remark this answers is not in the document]
jwerner-chromium: > code against this version should be able to interpret earlier versioned TLs

Yeah, that's just what I meant with "switch to code to interpret the respective version"... I was just trying to say that if the version number is older the implementation needs to treat it as that older version and e.g. not expect fields to be there that only exist in the version it was written for. I can try to rewrite that sentence to be more clear if I do another spin of this. Upgrading is also a potential option, I guess.

source/transfer_list.rst
danh-arm: Yes, you're (both) right it's not in general safe for old code to modify new TLs so I don't think we can change this policy in isolation. I wondered about using some bits of the version field to indicate to old code whether it's safe to modify the TL, but it seems unlikely that a new TL/TE header field will be added without affecting old writers in some way.

I'm still concerned about the interop problem but hopefully incrementing the version is sufficiently rare to make this not a problem in practice. I suppose old code could recreate the TL in the format it understands and hope that critical information is not lost when handing off to the next layer.

source/transfer_list.rst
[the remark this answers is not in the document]
danh-arm: OK, I guess this is quite a minor comment. If I have time I'll propose alternative wording here.

source/transfer_list.rst
jwerner-chromium: Sorry, upon re-reading this I don't know what I was thinking here. This makes no sense. The `alignment_offset` is necessary because that's the only way the old `tl_base_addr` gets into this calculation at all. The new code here doesn't handle TL's that are not themselves aligned to the maximum alignment requirement correctly.

Uploaded PR #12 to fix this and revert to the original version of this PR for this section.

PR #7, opened 2023-02-22
Clarify expectations for tag allocation process

jwerner-chromium:
This patch rewrites the section about entry type ranges to try to make it more clear how tags are supposed to be allocated and what standardized tags can be used for, and how non-standardized tags should be avoided to preserve interoperability. Also move the non-standardized range to the end of the tag space, as discussed during review.

### DIFF
--- a/source/transfer_list.rst
+++ b/source/transfer_list.rst
@@ -364,15 +364,57 @@ Inputs:
    to `tl.checksum` (`new_tl_base + 0x4`).
 
 
-Entry type ranges
------------------
+Entry type allocation
+---------------------
 
-The content of the data section is determined by the tag id. The tag id space contains two ranges:
+Tag IDs must be allocated in this specification before use. A new tag ID can be
+allocated by submitting a pull request to this repository that adds a
+description of the respective TE data layout to this specification. Tag IDs do
+not have to be allocated in order. Submitters are encouraged to try to group
+tag IDs together in logical clusters at 16 or 256-aligned boundaries (e.g. all
+tags related to a particular chipset or to a particular firmware project could
+use adjacent tag numbers), but there are no predefined ranges and no
+reservations of tag ranges for specific use.
 
- #. Standard range, and
- #. Non-standard range
+Tags are expected to have a simple layout (representable by a C structure) and
+each tag should only represent data for a single logical concept. Data for
+multiple distinct concepts should be split across different tags, even if
+they're always expected to appear together on the first platform adding the tag
+(to encourage reusability in different situations). Alternatively, complex data
+may be represented in a different kind of well-established handoff data
+structure (e.g. FDT [DT]_, HOB [PI]_) that is inserted into the TL as a single
+TE. The same tag ID may occur multiple times in the TL to represent multiple
+instances of the same kind of object. Tag layouts (including the meaning of all
+fields) are considered stable after being added to this specification and may
+never be changed in a backwards-incompatible way. If a backwards-incompatible
+change is desired, a new tag ID should be allocated for the new version of the
+layout instead.
 
-The `tag_id` ranges are described in :numref:`tab_tag_id_ranges`.
+Tag layouts may be changed in a backwards-compatible manner by allowing new
+valid values in existing fields (including reserved fields), as long as the
+original layout definition clearly defined how unknown values in those fields
+should be handled, and the rest of the TE would still be considered valid and
+correct for older readers that consider the new values unknown. TE layouts may
+also be expanded by adding new fields at the end, with the same restrictions.
+TEs should not contain explicit version numbers and instead just use the
+`data_size` value to infer how many fields exist. TE layouts which have been
+changed like this must clearly document which fields or valid values were added
+at a later time, and in what order.
+
+The {0xff_f000, ..., 0xff_ffff} range is reserved for non-standardized use.
+Anyone is free to use tags from that range for any custom TE layout without
+adding their definitions to this specification first. The use of this range is
+*strongly discouraged* for anything other than local experiments or code that
+will only ever be used in closed-source components owned by the entity
+controlling the entire final firmware image. In particular, a TE just
+containing platform-specific data or internal structures specific to a single
+firmware implementation is no reason not to allocate a standardized tag for it
+in this specification. Since standards often emerge organically, the goal is to
+create unique tag IDs for everything just in case it turns out to be useful in
+more applications than initially anticipated. Basically, whenever you're
+submitting code for a new TE layout to any public open-source project, that's
+probably a good indication that you should allocate a tag ID for it in this
+specification.
 
 .. _tab_tag_id_ranges:
 
@@ -382,27 +424,19 @@ The `tag_id` ranges are described in :numref:`tab_tag_id_ranges`.
    * - tag ID range
      - Description
 
-   * - 0x0 -- 0xf_ffff
-     - Standard tag id range. Any tag id in this range must first be allocated in this specification before being used. The allocation of the tag id requires the entry layout to be defined as well.
+   * - 0x0 -- 0x7f_ffff
+     - Standardized range. Any tag ID in this range must first be allocated in this specification before being used. The allocation of the tag ID requires the entry layout to be defined as well.
 
+   * - 0x80_0000 -- 0xff_efff
+     - Reserved. (Can later be used to extend standardized range if necessary.)
 
-   * - 0x10_0000 -- 0x10_ffff
-     - Non-standard range. A platform firmware integrator can create entries in this range. Different platforms are allowed to have tag ids in this range with distinct data formats. Entries in this range are not standardized.
-
-   * - 0x11_0000 -- 0xffff_ffff
-     - Reserved
+   * - 0xff_f000 -- 0xff_ffff
+     - Non-standardized range. Tag IDs in this range can be used without allocation in this specification. Using this range for anything other than local experimentation or closed-source components that are entirely under the control of a single platform firmware integrator is strongly discouraged. Tags in this range are not tracked in this repository and PRs to add tag defintions for this range will not be accepted.
 
 .. _sec_std_entries:
 
 Standard transfer entries
 -------------------------
-
-The TEs have a `tag_id` in the {0, ..., 0xf_ffff} set. Both
-the tag_id of a standard entry as well as the entry layout
-must be defined in this specification before being used.
-New entries are expected to have a simple layout. Complex
-data should be represented in a self-describing data
-structure, such as the FDT [DT]_.
 
 The following entry types are currently defined:
 
@@ -411,8 +445,6 @@ The following entry types are currently defined:
 - single HOB block entry: tag_id = 2 (:numref:`hob_block_entry`).
 - HOB list entry: tag_id = 3 (:numref:`hob_list_entry`).
 - ACPI table aggregate entry: tag_id = 4 (:numref:`acpi_aggr_entry`).
-
-All other standard `tag_id` values are reserved by this specification.
 
 .. _void_entry:
 


source/transfer_list.rst
```diff
@@ -164,28 +198,20 @@ The `tag_id` ranges are described in :numref:`tab_tag_id_ranges`.
...
 .. _sec_std_entries:
 
 Standard transfer entries
 -------------------------
 
-The TEs have a `tag_id` in the {0, ..., 0xf_ffff} set. Both
-the tag_id of a standard entry as well as the entry layout
-must be defined in this specification before being used.
-New entries are expected to have a simple layout. Complex
-data should be represented in a self-describing data
```
sjg20, line 186: Please keep this last sentence

source/transfer_list.rst
```diff
@@ -146,15 +146,49 @@ The TE header is defined in :numref:`tab_te_header`.
...
-
-The `tag_id` ranges are described in :numref:`tab_tag_id_ranges`.
+Entry type allocation
+---------------------
+
+Tag IDs must be allocated in this specification before use. A new tag ID can be
+allocated by submitting a pull request to this repository that adds a
+description of the respective TE data layout to this specification. Tag IDs do
+not have to be allocated in order. Submitters are encouraged to try to group
+tag IDs together in logical clusters at 256-aligned boundaries (e.g. all tags
```
sjg20, line 156: 256 seems like a lot. Perhaps this can be handled by adding a comment above the allocation which indicates how many are likely to be needed? For many projects, 16 might be enough.

I suggest "16-aligned (or if necessary 256-aligned) boundaries"

source/transfer_list.rst
```diff
@@ -146,15 +146,49 @@ The TE header is defined in :numref:`tab_te_header`.
...
+
+Tags are expected to have a simple layout (representable by a C structure) and
+each tag should only represent data for a single logical concept. Data for
+multiple distinct concepts should be split across different tags, even if
+they're always expected to appear together on the first platform adding the tag
+(to encourage reusability in different situations). The same tag ID may occur
+multiple times in the TL to represent multiple instances of the same kind of
+object. Tag layouts (including the meaning of all fields) are considered stable
+after being added to this specification and may never be changed in a backwards
+incompatible way. If a backwards-incompatible change is desired, a new tag ID
```
sjg20, line 169: move the 'backwards' and the end of 168 to the next line so you can say 'backwards-incompatible'

source/transfer_list.rst
```diff
@@ -146,15 +146,49 @@ The TE header is defined in :numref:`tab_te_header`.
...
+should be allocated for the new version of the layout instead.
+(Backwards-compatible ways to change existing layouts could be increasing the
+data_length to append new fields behind the structure or using fields in the
+layout that were previously marked as reserved, as long as the TE would still
+be considered valid for older implementations that ignore these new fields. TE
+layouts which have been changed like that must clearly document which fields
+have been added after the first version.)
```
sjg20, line 176: Should we prohibit a version number in the TE? It seems like that could be an invitation to do crazy things.

Also, should we suggest that people can allocate a new tag if the old one has outlived its usefulness?

source/transfer_list.rst
```diff
@@ -146,15 +146,49 @@ The TE header is defined in :numref:`tab_te_header`.
...
+
+The {0xff_f000, ..., 0xff_ffff} range is reserved for non-standardized use.
```
sjg20, line 178: Should we be explicit that PRs allocating things in this space will not be accepted?

sjg20 (changes requested):

source/transfer_list.rst
jmarinho: I agree that a version field should not be allowed in the TE definition (maybe we can police/enforce that during the TE PR review).
A newer consumer needs to still be able to detect the older layout.
I expect that to be done via the data_size field.
Are there cases where the data_size is ambiguous in identifying the v1 and v2 of a TE layout?

source/transfer_list.rst
```diff
@@ -164,28 +198,20 @@ The `tag_id` ranges are described in :numref:`tab_tag_id_ranges`.
    * - tag ID range
      - Description
 
-   * - 0x0 -- 0xf_ffff
-     - Standard tag id range. Any tag id in this range must first be allocated in this specification before being used. The allocation of the tag id requires the entry layout to be defined as well.
+   * - 0x0 -- 0x7f_ffff
+     - Standardized range. Any tag ID in this range must first be allocated in this specification before being used. The allocation of the tag ID requires the entry layout to be defined as well.
 
+   * - 0x80_0000 -- 0xff_efff
```
jmarinho, line 204: This has 4096 ids for the non-standard range.
Previously the 0x10_0000 -- 0x10_ffff had 65536.

4096 is more than enough ids for the non-standard range, and I'm OK with that!
Just wanted to point this out in case this was not intentional.

source/transfer_list.rst
jwerner-chromium: Well, I'm not really sure how to fit that single sentence into the rewrite. I also feel it is ill-defined... what is "complex data"? This is what I tried to describe more clearly in the second paragraph of my new text ("Tags are expected to have a simple layout (representable by a C structure) and each tag should only represent data for a single logical concept. ..."). Doesn't that already disallow "complex data"? If not please explain what kind of scenario you think shouldn't be allowed that you think isn't covered by my text and I'll try to figure out how to fit that in.

source/transfer_list.rst
jwerner-chromium: There are 32768 groups of 256 tags in the standardized range (and almost again that many if we start using the reserved range). Do you really think it's likely that we'll find that many independent stakeholders who want to add their own things here?

I really think 256 is the most convenient trade-off here -- it's large enough that few projects will likely need to allocate a second one, but still gives us so many independent blocks that I think it's extremely unlikely we'll run out (and, after all, in the extremely unlikely event we do run out we can still start reusing the high 128 bits of old blocks with just 3 entries where nothing new has been added in years... this is just guidance, not something that is forever set in stone). If anything I'd wonder if it should go up to 512 or 1024, but in the end it's not that much of a difference.

But I think 16 is so small that it will likely require most projects to allocate multiple fields and kinda defeat the "group similar stuff together" idea. I really don't think we need to be that overly frugal with this. 23 bits is a _lot_ of ID space and there's no prize waiting for us if 10 years from now 99% of it is still unused.

source/transfer_list.rst
jwerner-chromium: Done

source/transfer_list.rst
jwerner-chromium: Okay, I rewrote this part a bit to more explicitly define how backwards-compatible changes should be made, and include the version number prohibition. (In practice, of course, version numbers may still be included in cases where a separately defined data structure like an FDT or ChromeOS' vboot work buffer is embedded in a TE.)

source/transfer_list.rst
jwerner-chromium: Done

source/transfer_list.rst
jwerner-chromium: Right, no, that was intentional (I think it might have been a suggestion from Dan on the previous PR).

source/transfer_list.rst
jmarinho: I agree that this is mostly guidance and we can adapt as we go.
It's also true that 32768 groups of 256 tags is a lot. We'll realistically never run out of these groups.

jmarinho (dismissed):

source/transfer_list.rst
danh-arm: I remember suggesting to move the non-standard range to the end but I don't remember anything about the size. Anyway, this LGTM.

danh-arm (dismissed):

jwerner-chromium:
Rebased onto latest HEAD to resolve conflicts with my other PR that just landed, otherwise no changes since last week.

danh-arm (approved):

source/transfer_list.rst
sjg20: You can just leave it  as is, if you like.

But if you want to expand it...

Complex data is anything more than a simple C struct. So for example, ACPI and SMBIOS would be complex. Multiple C structs would be complex. Perhaps we should set a field limit as well, so people don't use it for FSP parameters.

How about:

Complex data, such as multiple C structs or structs with more than 5-10 fields, should be represented in a self-describing data structure, such as the FDT [DT]_

sjg20 (changes requested):

jmarinho (dismissed):

source/transfer_list.rst
jwerner-chromium: Sorry, I really don't understand the second part. Doesn't 0x101, 0x0102, 0x103 vs. 0x0201, 0x0202, 0x0203 make it much easier to see which tags are grouped together (0x1 vs. 0x2 is the group prefix and the last two hex digits are within the group)? In what way is that "harder" than being able to tell that 31 is in a different group than 32? I mean, most of us probably know the multiples of 16 pretty well but it's still easier to not have to do the math at all because the split is already built into the digits?

sjg20: Well I believe in practice we would end up with:

100, 101, 102
200
300, 301, 302, 303, 304, 305, 306, 307
400

etc.

which is using up a lot of space and making the numbers 10x larger than they need to be, just in case a project generates 100 (256 decimal) tags

I would rather have things closer together. I suppose we could allow projects to take 16 groups if they want to, then we get the best of both worlds

source/transfer_list.rst
jwerner-chromium: I don't think it makes sense to come up with a hard field limit, that's why I tried to explain this as "should represent single logical concept" instead (and the actual determination can then be made by reviewers on a case by case basis). For example, the framebuffer information structure coreboot passes to its payloads contains 14 distinct fields, but I don't think that means it would be too "complex" for this data structure... it's just defining the bit offset and size of each color channel in the packed pixel representation in an extensible way, and the cleanest way to do that is with a separate mask and offset field for each channel.

Not sure what exactly you mean by "multiple C structs". I think the case of having a simple array of a sub struct should be allowed, because it is very useful and comes up often, e.g. something like:
```
struct timestamp_table_tag {
  uint32_t num_entries;
  struct timestamp_entry {
    uint32_t entry_id;
    uint32_t entry_value;
  }[];
}
```
So I think that's roughly the limit of what should be allowed and I think that's covered by my text (it is still "representable by a C structure", i.e. my definition of `struct timestamp_table_tag` above is valid C code). A construct where, say, multiple structures are placed back-to-back and some of them are optional based on values set in the first structure would not be allowed because that's not "representable by a C structure" anymore. (Of course, wrapping a complex legacy structure in a tag as an opaque byte array is still allowed because that's what the HOB, FDT and ACPI tags are already doing, but that should not be done for new structures created exclusively for the transfer list.)

sjg20: I think it would faster to come to agreement if you respond to what people say, rather than inferring things that were not said. I dd not suggest a 'hard' field limit. 

A (soft) field limit is useful as it provides guidance as to when you are off the reservation. The example you show makes sense I suppose. What happens when you want to add more entries to the timestamp log? I suppose you create another struct with the same tag later on in the TL.

It seems we mostly agree with the intent. My goal is to actually express this in text, so please suggest an alternative to what I wrote.

source/transfer_list.rst
jwerner-chromium: Please see the calculations about how many groups of 256 tags can fit into the available space -- are you really concerned that this specification will draw tens of thousands of different firmware projects and SoC vendors? I mean, I am really hoping this will take off too, but that number still doesn't seem at all realistic to me. We don't gain anything from leaving all this ID space unused "just in case". I do however think that one consumer needing more than 16 tags is going to be pretty common (e.g. the handoff list from coreboot to its payload has 50+ tags at the moment).

> I would rather have things closer together. I suppose we could allow projects to take 16 groups if they want to, then we get the best of both worlds

If we should actually run out of groups (which I think is astronomically unlikely with 32K), we can still start reusing space from those that haven't been filled up much. Remember, this is meant to be a guideline, not a hard lease. We're not giving those 256 tags away to someone else who will never give them back, we're just saying "please pick a start number on this alignment and try to group all your stuff together there for as long as it fits".

I don't understand what the benefit of "having things closer together" is supposed to be. The whole point is that all groups are independent and there is no order or relationship between different groups.

sjg20: I am not worried about the amount of space. I'm just trying to avoid spreading out a lot over that space.

Another option would be to just allocate things sequentially and not worry about having each project's numbers next to each other. Would that work?

source/transfer_list.rst
sjg20: 1. I'd like to see guidance on the number of members in the structs, e.g. no more than 5-10 fields. It will save us arguing about it in PR reviews
2. The bit about complex structures using FDT needs to stay as it is important to explain the fallback

jwerner-chromium: Well, I really disagree with calling out any specific number of fields, because as I have tried to explain and illustrate with examples earlier, this metric is really meaningless when some things that belong together just require a bunch of fields to describe while on the other hand I can easily write a TE that only contains 3 fields but they all concern completely independent things that should be described in an independent TE each. We need a more abstract rule to describe this restriction.

We have discussed the issue with the FDT comment to death already. This is supposed to be a universal handoff specification with guidelines about how things should be represented but no restrictions on how projects are allowed to use it or what they are allowed to represent. This is not just supposed to be a pre-RAM substitute for FDT. It should not tell users to use any specific other data structure instead, it should explain how to properly represent any kind of information in this structure.

Earlier you commented "You can just leave it as is, if you like.", but it sounds like you are retracting that now? @jmarinho , @danh-arm , do you have opinions on this topic that might help us resolve this impasse?

source/transfer_list.rst
jwerner-chromium: > I'm just trying to avoid spreading out a lot over that space.

I really don't understand... _why?_ What's the problem with spreading out? It's an ID space! It doesn't matter where in there you allocate IDs, other than for aesthetics and to more easily recognize patterns when reading raw IDs in a hex dump.

> Another option would be to just allocate things sequentially and not worry about having each project's numbers next to each other. Would that work?

Yes, it would, it would just look a lot more messy. I really don't understand why you're preferring this solution. I think grouping has a clear aesthetics/recognizability benefit, and it may also reduce collision churn when many new ID PRs are submitted and reviewed at once. What's the benefit of not "speading out a lot over that space" that outweighs these?

source/transfer_list.rst
danh-arm: I've been avoiding commenting here because it feels to me like we're getting into bike shedding territory. I don't feel strongly about this; both Jose and I have approved this version so we could merge it now, although it would be nice to have general consensus. For what it's worth, my opinion is that I agree with Julius that guidance on the max number of fields seems unnecessarily restrictive. At the same time, I think the existing words to use something like FDT for complex types is OK. I think I suggested in a previous review that maybe this wording could be softened to achieve consensus, e.g. "For complex data, consider the use of a self-describing data structure, for example FDT [or HOB or ACPI]".

source/transfer_list.rst
danh-arm: Again, I don't feel strongly here. FWIW I also don't understand the issue with spreading out the tag ID space. The compiler efficiency argument doesn't seem strong to me (assuming it's valid at all).

sjg20: I am OK with clumping them together, but I think clumps of 16 is enough. The compiler argument is valid on 32-bit ARM at least, where small numbers can be set in a register. But in general, smaller numbers are easier to read for humans. For example, 46 is easier that 106 just from the smaller number of digits. One of the things I hate about UUIDs is that they are so long. So why make things longer than necessary, just to keep the computer happy? These tags are for humans and we will be seeing them in binary dumps for years.

So that is the 'why'.

Most likely we'll have 5-6 projects using this initially, with fewer than 100 tags in use. Keeping these somewhat together seems best to me.

jmarinho: If I'm to give my opinion, we're discussing a guidance statement that we'll have the liberty to adapt as we go along.

I see the point of both 16 and 256 groupings.
It is best to have tag ids that are more human-readable. On the other hand having 256 groupings does give more room for each "cluster" to grow.
Is there a compromise value? Say 64 perhaps?

I don't think that this discussion should gate us from a 0.9 release, and starting to work on the implementations.

jwerner-chromium: > So why make things longer than necessary, just to keep the computer happy? These tags are for humans and we will be seeing them in binary dumps for years.

It's usability for humans that I'm trying to argue for here! For the computer this is all the same. I find it very contrived to argue that there was any difference in readability between 46 and 106, I think for anything under 5 digits it really doesn't make a difference to human readers. But it does make a notable difference whether you can say "oh look, this is a 0x03xx ID, that means it must be one of the U-Boot tags" or you say "gee, I can't remember, was 0x024x one of the 3 coreboot prefixes or maybe one of the 4 U-Boot ones...". I don't think it's fair to compare an (at _most_, if this spec is far more successful than we could hope for) 6-digit hex number to a UUID in terms of human readability.

> Is there a compromise value? Say 64 perhaps?

I think that would be losing the readability advantage of breaking the ID at a hex boundary.

source/transfer_list.rst
jwerner-chromium: I really feel that this spec should contain no suggestions of the form "for XXX just use <something else> instead". I think a universal firmware handoff specification should be able to serve all firmware handoff needs. Otherwise, what's the point in trying to make something new that's focused on interoperability if half the stuff you might want to interoperate on is not covered by it and has to be represented by something else that requires some other parser again? We should not be precluding anyone from mixing this with other data structures if they want to but I don't think we should be suggesting it as the standard solution either. Those projects that want to live purely based on this specification should be able to do so without needing a fallback for specific kinds of data.

I fully understand the "tag layouts should not be as complex as an ACPI table" argument, and I've been trying to write guidelines in this PR that should rule that out (I just don't think specifying it in number of fields is a good approach). But I think the correct solution to the "complex data" problem is to break it down into multiple pieces of simple data that are passed as multiple TEs. If the assertion here is that there exist some kinds of data which firmware might want to pass between stages that could never possibly be broken down into a form that would be appropriate for the TL and will always require wrapping in some other data structure, then I fundamentally disagree with that. This spec should be able to stand on its own and offer at least one correct option to solve every firmware handoff use case by itself.

source/transfer_list.rst
sjg20: That's all fine as a differing opinion, but I have answered your questions with a reasonable response.

Going back to my original point, how many are likely to be needed? For many projects, 16 might be enough? You didn't address the efficiency angle. Also I am a reader and I prefer shorted numbers to longer ones.

If we cannot agree this here, then I propose dropping all mention of how they will be allocated. 

We can also do it on the fly...e.g. if coreboot needs a range of 0x100 tags, it can have its own area allocated. For projects that don't, they can use a smaller area of 0x10.

source/transfer_list.rst
danh-arm: > But I think the correct solution to the "complex data" problem is to break it down into multiple pieces of simple data...

That is one valid solution but I don't think that should be the goal of this specification. In systems that have a normal world bootloader operating on data in a particular format (FDT, HOB, etc..) and has earlier firmware operating on the same data, it _may_ be a better design choice to keep the data in that format throughout, rather than doing conversions.

This isn't worth arguing about, though. Let's leave the sentence out - it isn't of material importance.

source/transfer_list.rst
danh-arm: I'm not keen on dropping all mention of how they will be allocated but I agree we don't necessarily need a single block size. Could we just say to try to group in clusters of 16 or 256 aligned boundaries, depending on the likely number of TEs required?

source/transfer_list.rst
jwerner-chromium: Right, I in no way want to preclude systems where that makes sense to use it like this (and I think the fact that the FDT TE is one of the first defined in this spec makes it clear that wrapping those formats is a valid option). But if there's a sentence that just says "complex data structures need to be represented in an FDT", that's not an option, that's a requirement. I don't want this to later be used in reviews to force an FDT dependency onto projects that don't want to have one.

Would it work as a compromise if I add a new sentence like this: "Data for multiple distinct concepts should be split across different tags, [...]. _Alternatively, complex data may be represented in a different kind of well-established handoff data structure (e.g. FDT, ACPI table) that is inserted into the TL as a single TE._"?

danh-arm: Yes, your new sentence works for me!

source/transfer_list.rst
sjg20: Yes that sounds good to me, using 16 or 256 depending on likely number ot TEs.

source/transfer_list.rst
jwerner-chromium: Updated (replaced "ACPI table" with "HOB" in the example).

source/transfer_list.rst
jwerner-chromium: Updated the text although I am concerned about the logistics of this (e.g. if one project starts grabbing tags at 0x200, 0x201, 0x202... how will the next project know whether they should continue at 0x210 or 0x300?). Guess we can argue about that in the individual submission reviews later.

source/transfer_list.rst
```diff
@@ -364,15 +364,57 @@ Inputs:
...
+use adjacent tag numbers), but there are no predefined ranges and no
+reservations of tag ranges for specific use.
+
+Tags are expected to have a simple layout (representable by a C structure) and
+each tag should only represent data for a single logical concept. Data for
+multiple distinct concepts should be split across different tags, even if
+they're always expected to appear together on the first platform adding the tag
+(to encourage reusability in different situations). Alternatively, complex data
+may be represented in a different kind of well-established handoff data
+structure (e.g. FDT, HOB) that is inserted into the TL as a single TE. The same
```
sjg20, line 385: For FDT can you please keep the link to the spec? Also you've replaced 'self-describing' with 'well-established' which is not the same thing. Can you put that back?

HOB would be very confusing since it includes pointers and is tied to use of EFI...so I suggest just dropping that and saying (e.g. FDT)

sjg20 (changes requested):

source/transfer_list.rst
jwerner-chromium: I don't think this specification should be passing judgement on which other handoff formats are "good" or "bad". Different projects may be tied to different legacy structures that regardless of what disadvantages they may have still need to be used for certain reasons. The point here is to say "if you want to describe stuff in the transfer list directly, follow these rules -- otherwise, you can also wrap your existing legacy strucutre in a TE and keep using that for some data". I am not very familiar with HOB and am not trying to champion it specifically but I think this guidance should be solution-agnostic, so it would be good to list more than one example to illustrate that. Since there's already a TE defined for HOBs they are clearly already allowed in this role?

I don't really understand what "self-describing" is supposed to mean in this context, or why it is needed. What makes the FDT self-describing in a matter that other structures are not? (Are HOBs "self-describing"?) An FDT is still a binary (i.e. not all-text) structure that I need a parser to read, after all. Do you just mean "a key-value store with ASCII keys"? Requiring that here seems unnecessarily narrow to me.

danh-arm: I agree it would be good to have spec links but I think this is a minor comment. Otherwise looks good to me. We know of at least 1 partner who wants to use HOBs in the TL. I tend to agree the self-describing comment doesn't add much value; I think this just refers to the fact that the structure contains a header with a signature. In that sense, the TL (and all specified TEs) are self-describing, even if individual TEs are not.

danh-arm (dismissed):

source/transfer_list.rst
sjg20: We may have forgotten why this was there.

The first thing that came up for "self-describing data structure" just now was "A message that contains data as well as the metadata that describes the format and the meaning (i.e., the syntax and the semantics) of that data. For example, Extensible Markup Language (XML) is a self-describing message format that consists of tag/value pairs."

Clearly a HOB Is not self-describing. You must have the format to decode the data. In fact even that might be hard. For example, each HOB entry has a GUID which indicates its type, but these can be private and unknown.

The goal here is, as I understand it, to avoid people creating a new binary structure that cannot be read by other projects, except for pre-existing formats, i.e. HOB and ACPI. The reason for mentioning FDT is that it can be read without needing a secret spec or code, etc.

We already have HOB list in the spec though, as a tag type:

https://github.com/FirmwareHandoff/firmware_handoff/blob/main/source/transfer_list.rst#hob-list-entry-layout-xferlist_hob_l

I am OK with submitting this without the link to the FDT spec if that helps. I can do a fix-up PR later.

So, please, can we have something like:

Alternatively, complex data may be represented in a self-describing data format such as the FDT [DT]_.

danh-arm: > The goal here is, as I understand it, to avoid people creating a new binary structure that cannot be read by other projects, except for pre-existing formats, i.e. HOB and ACPI

Hmm, I thought the goal was simply to avoid people creating a new binary structure when a suitable one already exists. Although HOB/ACPI are not self-describing, I don't think that means very much in this context because all binary structures become self-describing when wrapped in a TE specified in this spec. So HOB is a reasonable choice for some implementations and worthy of mention IMO.

jwerner-chromium: > The reason for mentioning FDT is that it can be read without needing a secret spec or code, etc.

Sorry, but this is just completely untrue. The FDT is a binary structure and if you don't have the right structure definitions for decoding the FDT header, node structure, property structure, reserve list, string table and all that gunk, it is just as much of an opaque blob to you as a HOB. That is my fundamental objection to making FDT stand out like it was something special and "better" here, because it isn't. It's a hierarchical key-value store with ASCII keys and integer-array or ASCII values. HOB lists are a key-value store with GUID keys and memory buffer values. (The TL itself, for comparison, is a key-value store with integer keys and C structure values.) Those are not the same thing, but I don't see how the difference matters enough for this context to the point where we need to tell people to prefer one over the other. If they want to be interoperable, they should be using TEs directly anyway, and if they want to stick to whatever other legacy structures their stack is already used to, they should be free do that too.

An XML document is different because it includes a link to a schema doc that describes the meaning of the fields (which an FDT has not), so I assume that's why wherever you pulled that quote from calls it "self-describing" (not that I think that this property was very useful for our purposes here either).

sjg20: I won't respond to the first paragraph other than to say FDT is considerably less opaque than a binary blob.

Dan, the point of this text was to avoid people creating a complex new binary format and adding it to the TL. Those days should be behind us, although we need to support legacy formats in some cases.

Since you both want to mention HOB, that's fine. But please add the FDT link back.

jwerner-chromium: Here's a hexdump of a random FDT, let's see how opaque it is:
```
00000000: d00d feed 0001 69a0 0000 0038 0001 5a84  ......i....8..Z.
00000010: 0000 0028 0000 0011 0000 0010 0000 0000  ...(............
00000020: 0000 0f1c 0001 5a4c 0000 0000 0000 0000  ......ZL........
[...]
00001080: 0000 0003 0000 0008 0000 01c5 0000 000a  ................
00001090: 0000 0000 0000 0003 0000 0004 0000 00f9  ................
000010a0: 0000 001a 0000 0001 6c32 2d63 6163 6865  ........l2-cache
000010b0: 0000 0000 0000 0003 0000 0006 0000 0032  ...............2
000010c0: 6361 6368 6500 0000 0000 0003 0000 0004  cache...........
000010d0: 0000 01a5 0000 000b 0000 0003 0000 0004  ................
000010e0: 0000 00f9 0000 000f 0000 0002 0000 0002  ................
000010f0: 0000 0001 6370 7540 3530 3000 0000 0003  ....cpu@500.....
00001100: 0000 0004 0000 012c 6370 7500 0000 0003  .......,cpu.....
[...]
00015eb0: 6c2d 6e61 6d65 7300 7069 6e63 7472 6c2d  l-names.pinctrl-
00015ec0: 3000 7069 6e63 7472 6c2d 3100 766d 6d63  0.pinctrl-1.vmmc
00015ed0: 2d73 7570 706c 7900 7671 6d6d 632d 7375  -supply.vqmmc-su
00015ee0: 7070 6c79 0072 6571 7569 7265 642d 6f70  pply.required-op
00015ef0: 7073 006f 7070 2d61 7667 2d6b 4270 7300  ps.opp-avg-kBps.
00015f00: 6373 2d67 7069 6f73 0073 7069 2d6d 6178  cs-gpios.spi-max
00015f10: 2d66 7265 7175 656e 6379 0067 7069 6f2d  -frequency.gpio-
[...]
```
Certainly completely different from a binary blob!

Updated to add links for both examples.

sjg20 (approved):
I'm OK with this so long as I can re-add the 'self-describing' part later. I think this has been going on for too long.

source/transfer_list.rst
sjg20: You could try fdtdump ?

jwerner-chromium: Yes, and I can write dump tools for any other imaginable data structure too. I still fail to see what inherently makes the FDT "self-describing" here.

sjg20: You can see what fields are present and read the values. With a binary blob you have no idea what the data means.

You can't write a dump tool if you don't know the format of the data. FDT forces that into the open, at least to an extent.

It is not fully self-describing, since the schema is not included. But even without that, it helps a lot.

jmarinho (approved):

jwerner-chromium:
Hi @jmarinho @danh-arm,

Just to clarify, there's nothing else you need from me for this PR, right? My understanding is that it is approved now and the remaining CI problem seems to be some infrastructure issue that I hope you can take care of. Let me know if I'm wrong and you're still waiting on something from me.

PR #9, opened 2023-05-30
Add GUID BLOB entry

mathias-arm:
This is a proposal to add an entry type for payloads identified by a GUID.


mathias-arm:
Folding this PR into #10 since the discussion is happening there.

PR #10, opened 2023-05-30
Add GUID POINTER entry

mathias-arm:
This is a proposal related to #9. It adds an entry type for payloads identified by a GUID, but instead of adding the content in the transfer list, it has a pointer to the content.


source/transfer_list.rst
```diff
@@ -657,3 +658,59 @@ such that the last ACPI table in this entry ends at offset
...
+          uint16_t data2;
+          uint16_t data3;
+          uint8_t data4[8];
+        } guid;
+
+        typedef struct  {
+          // GUID name
+          guid name;
+          // BLOB content
+          uint8_t blob[n];
```
sjg20, line 714: Can you please add the fields for this? What does it contain?

source/transfer_list.rst
```diff
@@ -714,3 +715,56 @@ semantics should be consistent (but not limited to) with [UEFI]_ and [ACPI]_.
...
+          uint32_t data1;
+          uint16_t data2;
+          uint16_t data3;
+          uint8_t data4[8];
+        } guid;
+
+        typedef struct {
+          guid name;          // GUID name
+          uint64_t address;   // content address
+          uint64_t size;      // content size
```
sjg20, line 769: What is the goal of putting this in the transfer list? We are trying to make the data visible and documented

sjg20 (changes requested):

source/transfer_list.rst
```diff
@@ -657,3 +658,59 @@ such that the last ACPI table in this entry ends at offset
...
+     - The size of `GuidBlob` (including BLOB content) in bytes.
+
+   * - guid_blob
+     - data_size
+     - hdr_size
+     - The guid_blob field contains a `GuidBlob`.
+
+.. code-block:: c
+
+        typedef struct {
```
jmarinho, line 703: nit: C-style datastructure may lead to ambiguous datastrcuture alignment and size.
To clarify the alignment and field sizes, does it make sense to represent this as a list of fields with size and offset like we did on the other entries?

mathias-arm:
Folding #9 in this PR.

source/transfer_list.rst
mathias-arm: I folded the content into the table.

source/transfer_list.rst
mathias-arm: I put a more detailed rationale in the description. The pointer variant would be useful if you want to keep an object at a specific location in memory. It's also an opportunity for optimizations: avoid moving content memory, avoid `XFERLIST_VOID` some entries due to alignment, etc.

source/transfer_list.rst
mathias-arm: I folded the C-like description into their corresponding tables.

source/transfer_list.rst
mathias-arm: I am not sure I understood the part about "visible and documented" w.r.t. this part of the PR.

mathias-arm: Let me know if the updates did not address your question.

sjg20:
I believe it is better to actually add tags and a description for each structure you want to add to the TH.

Adding an anonymous GUID and some data goes against the spirit of this.

Can you give a specific example of a TE you want to add and what it contains?

We did talk about having a way to add a GUID into the TE header so it can be provided as well, but it is not in there yet.

danh-arm:
I agree with Simon that we need a specific use-case for this kind of TE. We don't want to speculatively create TEs "just in case". I don't think avoiding a large tag_id table is a good reason.
Also, an earlier version of this spec considered tags with pointers in it but that was discarded (at least for now) I think because of complications with the ensuring the pointed-to data is mapped correctly by both sender and receiver. It's not clear if this would be mapped separately to the TL itself. The spec already caters for adding TEs with alignment requirements (see the alignment field in the TL header), without having to relocate the TL. What is deficient in the current spec?

mathias-arm:
My use-case is that I want to add multiple UEFI objects (files, logs) in the transfer list so that the next firmware is not required to have access to a storage device.

I think my proposals are consistent with other entries, and/or your feedback is inconsistent with existing specifications. Applying the same logic/spirit you expressed, the `XFERLIST_HOB_B` and `XFERLIST_ACPI_AGGR` would need to be split into multiple different tags, i.e. a different tag for each `EFI_HOB_TYPE_*` value and ACPI table type. Otherwise, can you think of a way we could amend the descriptions in my proposal to restrict the scope of objects?

Two other feedback items are more specific to `XFERLIST_GUID_POINTER`:
- Regarding the issue of pointers in the transfer list, there are several HOB entries that have pointers in them (Firmware Volume HOB types, UEFI Capsule HOB, etc.), which, I think, would raise the mapping questions mentioned. This highlights the value of `XFERLIST_GUID_BLOB` as a transfer list container for memory-mapped addresses in HOB entries.
- For the alignment requirements, the current mechanism can work, but it is wasteful. If you want to have multiple page-aligned entries, you need to insert one page between each entry, mosly occupied by an empty entry and the transfer entry header.

sjg20:
> My use-case is that I want to add multiple UEFI objects (files, logs) in the transfer list so that the next firmware is not required to have access to a storage device.

That sounds good and useful.

> 
> I think my proposals are consistent with other entries, and/or your feedback is inconsistent with existing specifications. Applying the same logic/spirit you expressed, the `XFERLIST_HOB_B` and `XFERLIST_ACPI_AGGR` would need to be split into multiple different tags, i.e. a different tag for each `EFI_HOB_TYPE_*` value and ACPI table type. Otherwise, can you think of a way we could amend the descriptions in my proposal to restrict the scope of objects?

For each item (files and logs), define a format and add it to the spec. Just start with one, say a log or a single file. It is hard to comment on things without more info.

> 
> Two other feedback items are more specific to `XFERLIST_GUID_POINTER`:
> 
> * Regarding the issue of pointers in the transfer list, there are several HOB entries that have pointers in them (Firmware Volume HOB types, UEFI Capsule HOB, etc.), which, I think, would raise the mapping questions mentioned. This highlights the value of `XFERLIST_GUID_BLOB` as a transfer list container for memory-mapped addresses in HOB entries.

But I think you should start by picking one and defining a struct for it.

> * For the alignment requirements, the current mechanism can work, but it is wasteful. If you want to have multiple page-aligned entries, you need to insert one page between each entry, mosly occupied by an empty entry and the transfer entry header.

What would be better?

danh-arm:
> My use-case is that I want to add multiple UEFI objects (files, logs) in the transfer list so that the next firmware is not required to have access to a storage device.

OK, so could you construct a TL with multiple `XFERLIST_HOB_B`s of type  _EFI_HOB_GUID_TYPE? The specific HOB types don't need to be specified (again) in this specification.

I think you should either be defining TEs for the specific structures you want to pass (as Simon suggests) or just use the existing container formats (e.g. HOB, DT or ACPI) as appropriate. I don't see what value there is in defining TLs for intermediate structures, like GUID BLOBs.

> Regarding the issue of pointers in the transfer list, there are several HOB entries that have pointers in them...

Right, but that is a contract between HOB producers and consumers, which doesn't need to be part of this spec.

sjg20:
Can we delete the PR?

PR #16, opened 2023-08-24
Add new entry for TPM info

tusharkhandelwal2:
This entry will have information related to TPM event log and CRB buffer.

### DIFF
--- a/source/transfer_list.rst
+++ b/source/transfer_list.rst
@@ -486,7 +486,9 @@ The following entry types are currently defined:
 - single HOB block entry: tag_id = 2 (:numref:`hob_block_entry`).
 - HOB list entry: tag_id = 3 (:numref:`hob_list_entry`).
 - ACPI table aggregate entry: tag_id = 4 (:numref:`acpi_aggr_entry`).
-- Entries related to Trusted Firmware (:numref:`tf_entries`)
+- TPM event log entry: tag_id = 5 (:numref:`tpm_evlog_entry`).
+- TPM CRB base entry: tag_id = 6 (:numref:`tpm_crb_base_entry`).
+- Entries related to Trusted Firmware (:numref:`tf_entries`).
 
 .. _void_entry:
 
@@ -696,6 +698,105 @@ such that the last ACPI table in this entry ends at offset
      - data_size
      - hdr_size
      - One or more ACPI tables.
+
+
+.. _tpm_evlog_entry:
+
+TPM event log table entry layout (XFERLIST_EVLOG)
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+This entry type holds TPM-related information for a platform. The TPM event log
+info is a region containing a TPM event log as defined by TCG EFI Protocol
+Specification [TCG_EFI]_.
+
+.. _tab_tpm_evlog:
+.. list-table:: TPM event log type layout
+   :widths: 2 2 4 8
+
+   * - Field
+     - Size (bytes)
+     - Offset (bytes)
+     - Description
+
+   * - tag_id
+     - 0x3
+     - 0x0
+     - The tag_id field must be set to **5**.
+
+   * - hdr_size
+     - 0x1
+     - 0x3
+     - |hdr_size_desc|
+
+   * - data_size
+     - 0x4
+     - 0x4
+     - The size of the event log in bytes + sizeof(flags) i.e. 0x4.
+
+   * - flags
+     - 0x4
+     - hdr_size
+     - flags are intended to signal properties of this TE. Bit 0 is
+       need_to_replay flag. Some firmware components may compute measurements
+       to be extended into a TPM and add them to the TPM event log, but those
+       components are unable  to access the TPM themselves. In this case, the
+       component should set the "need_to_replay" flag so that the next
+       component in the boot chain is aware that the PCRs have not been
+       extended. A component with access to the TPM would replay the event log
+       by reading each measurement recorded and extending it into the TPM. Once
+       the measurements are extended into the TPM, then the "need_to_replay"
+       flag must be cleared if the transfer list is passed to additional
+       firmware components. Default value is "0". Other bits should be set to
+       zero.
+   
+   * - event_log
+     - data_size - 0x4
+     - hdr_size + 0x4
+     - Holds a complete event log.
+
+
+.. _tpm_crb_base_entry:
+
+TPM CRB base address table entry layout (XFERLIST_TPM_CRB_BASE)
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+The CRB info defines the address of a region of memory that has been carved out
+and reserved for use as a TPM Command Response Buffer interface.
+
+.. _tab_tpm_crb_base:
+.. list-table:: TPM CRB base type layout
+   :widths: 4 2 4 8
+
+   * - Field
+     - Size (bytes)
+     - Offset (bytes)
+     - Description
+
+   * - tag_id
+     - 0x3
+     - 0x0
+     - The tag_id field must be set to **6**.
+
+   * - hdr_size
+     - 0x1
+     - 0x3
+     - |hdr_size_desc|
+
+   * - data_size
+     - 0x4
+     - 0x4
+     - This value should be set to **0xc** i.e. sizeof(crb_base_address) + sizeof(crb_size).
+
+   * - crb_base_address
+     - 0x8
+     - hdr_size
+     - The physical base address of a region of memory reserved for use as a
+       TPM's Command Response Buffer region.
+
+   * - crb_size 
+     - 0x4
+     - hdr_size + 0x8
+     - Size of CRB.
+
+
 
 .. _tf_entries:
 
--- a/source/references.rst
+++ b/source/references.rst
@@ -12,3 +12,5 @@
 .. [OPTEECore] `OP-TEE Core <https://optee.readthedocs.io/en/latest/architecture/core.html>`
 
 .. [TFAFFAMB] `TF-A Secure Partition Manager: FF-A manifest binding to device tree <https://trustedfirmware-a.readthedocs.io/en/latest/components/ffa-manifest-binding.html>`
+
+.. [TCG_EFI] `https://trustedcomputinggroup.org/resource/tcg-efi-protocol-specification`.


source/transfer_list.rst
```diff
@@ -657,5 +658,58 @@ such that the last ACPI table in this entry ends at offset
...
+
+   * - evlog_base_address
+     - 8
+     - 8
+     - The base address of event log
+
+   * - evlog_size
+     - 0x4
+     - 16
+     - Size of the event log
```
sjg20, line 701: What does the event log contain? Can you provide the file format?

Otherwise this LGTM

sjg20 (changes requested):

source/transfer_list.rst
tusharkhandelwal2: Typically event log shows information about measurements taken during boot and has values extended to PCR registers. Event log is typically a binary file and follows TCG specifications.

apalos: @tusharkhandelwal2 wbefore merging this I think we should discuss what we need the EventLog for. 

There are 2 cases that come in mind. 
1. A previous stage boot loader e.g. TF-A constructed the EvenLog but doesn't have a TPM driver.  So the PCRs of the actual hardware were never extended.  In that case a bootloader which has the drivers needed will have to replay the EventLog on the hardware
2. The aforementioned bootloader can extend the PCRs normally, so the only thing we need is the EventLog base addr + size. 

So would it make sense to add a boolean describing if a bootloader needs the replay the EventLog?
We can probably derive that by reading PCR0 (which is required to be extended by the TCG specs), but I am not sure everyone will follow that spec.

source/transfer_list.rst
```diff
@@ -657,5 +658,58 @@ such that the last ACPI table in this entry ends at offset
...
+
+   * - crb_base_address
```
apalos, line 703: Why exactly do we need the CRB here?

source/transfer_list.rst
```diff
@@ -657,5 +658,58 @@ such that the last ACPI table in this entry ends at offset
...
+     - 8
+     - 8
```
jwerner-chromium, line 705: These offset numbers don't make sense, you can't have both the `evlog_base_address` and `crb_base_address` at offset 8. (Also, I think you should write all offsets in hex like the other entries.)

When you correct the offsets I would recommend to make sure all fields are naturally aligned, so since you have a couple of 8-byte and 4-byte fields you probably want to arrange them evlog_base_address, crb_base_address, evlog_size, crb_size.

source/transfer_list.rst
```diff
@@ -657,5 +658,58 @@ such that the last ACPI table in this entry ends at offset
...
+
+   * - hdr_size
+     - 0x1
+     - 0x3
+     - |hdr_size_desc|
+
+   * - data_size
+     - 0x4
+     - 0x4
+     - sizeof (base_address) + sizeof (evlog_size) + sizeof (crb_base_address) + sizeof (crb_size)
```
jwerner-chromium, line 691: Since the size of this entry must be constant I would recommend writing it like the tag_id description (i.e. "The data_size field must be set to **24**.")

source/transfer_list.rst
stuyod01: @apalos The CRB is a region of memory that a TPM implementation (e.g. TPM service in TrustZone) uses to communicate with TPM clients.  Firmware such as TF-A may need to carve out and reserve a region of memory for this purpose.  The "crb_base_address" field in the transfer list is so that firmware like TF-A can communicate the address/size of the carve out region to subsequent firmware components.

source/transfer_list.rst
stuyod01: I would not recommend using a directly link to a specific PDF.  Instead use the more general URL so that if the EFI spec has a minor rev the link will still be valid:
https://trustedcomputinggroup.org/resource/tcg-efi-protocol-specification/

stuyod01: > @tusharkhandelwal2 before merging this I think we should discuss what we need the EventLog for.
> 
> There are 2 cases that come in mind.
> 
> 1. A previous stage boot loader e.g. TF-A constructed the EvenLog but doesn't have a TPM driver.  So the PCRs of the actual hardware were never extended.  In that case, a bootloader that has the drivers needed will have to replay the EventLog on the hardware
> 2. The aforementioned bootloader can extend the PCRs normally, so the only thing we need is the EventLog base addr + size.
> 
> So would it make sense to add a boolean describing if a bootloader needs the replay the EventLog? We can probably derive that by reading PCR0 (which is required to be extended by the TCG specs), but I am not sure everyone will follow that spec.

I think adding a "need_to_replay" boolean would be useful.  It should default to zero.

source/transfer_list.rst
```diff
@@ -657,5 +658,58 @@ such that the last ACPI table in this entry ends at offset
...
+
+   * - evlog_base_address
+     - 8
+     - 8
+     - The base address of event log
```
stuyod01, line 696: The physical address of the region that contains the TPM event log.

source/transfer_list.rst
```diff
@@ -657,5 +658,58 @@ such that the last ACPI table in this entry ends at offset
...
      - One or more ACPI tables.
 
+
+.. _tpm_evlog_entry:
+
+TPM info table entry layout (XFERLIST_TPM_INFO)
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+This entry type holds TPM info of the platform.
+
```
stuyod01, line 668: I think adding a bit more information here would help provide context to readers.
Suggestion:

> This entry type holds TPM-related information for a platform.  The TPM event log info is a region containing a TPM event log as defined by TCG EFI Protocol Specification (https://trustedcomputinggroup.org/resource/tcg-efi-protocol-specification/).  The CRB info defines the address of a region of memory that has been carved out and reserved for use as a TPM Command Response Buffer interface.

source/transfer_list.rst
stuyod01: Agree with defining naturally aligned fields.  If we add a new "need_to_replay" boolean as described in an earlier comment we could use that to align the CRB info.

source/transfer_list.rst
apalos: Ok fair enough

source/transfer_list.rst
```diff
@@ -657,5 +658,65 @@ such that the last ACPI table in this entry ends at offset
...
+   * - hdr_size
+     - 0x1
+     - 0x3
+     - |hdr_size_desc|
+
+   * - data_size
+     - 0x4
+     - 0x4
+     - 0x18
```
stuyod01, line 694: The data size should be 0x19, given that we now have the need_to_replay field.

source/transfer_list.rst
```diff
@@ -657,5 +658,65 @@ such that the last ACPI table in this entry ends at offset
...
+     - Size of the event log
+
+   * - crb_size 
+     - 0x4
+     - 0x1C
+     - Size of CRB. crb_size == 0 means that crb_base_address is not valid
+
+   * - need_to_replay
```
jwerner-chromium, line 716: nit: I would consider calling this "flags" and defining bit 0 as the need_to_replay flag, with other bits reserved for future use.

source/transfer_list.rst
```diff
@@ -657,5 +658,65 @@ such that the last ACPI table in this entry ends at offset
...
+     - 0x1
+     - 0x20
+     - Value "1" means bootloader needs to replay the EventLog. Default value is "0".
```
jwerner-chromium, line 719: nit: Does the TPM specification define what exactly is meant by "replay"? I assume this means that the events stored in the log need to be written out to the TPM? This explanation could maybe use a little more detail, unless that is clear from context for everyone familiar with this specification.

Also, you might want to explicitly remind people here that the bootloader component that is doing this replay should update the TE to switch the flag from 1 to 0, so that a later bootloader component won't think it needs to be replayed again.

jwerner-chromium (approved):
Looks good to me, a few nits

source/transfer_list.rst
tusharkhandelwal2: I think this can be updated in the Description part.

source/transfer_list.rst
```diff
@@ -657,5 +658,65 @@ such that the last ACPI table in this entry ends at offset
...
+
+   * - evlog_base_address
+     - 0x8
+     - 0x8
+     - The physical address of the region that contains the TPM event log.
+
+   * - crb_base_address
+     - 0x8
+     - 0x10
+     - The base address of CRB.
```
sjg20, line 704: We should not be putting addresses in there...instead the contents of that region should be included in the TE

sjg20 (changes requested):

source/transfer_list.rst
stuyod01: This field needs to contain the physical address of the CRB.

Suggested re-wording:

> The physical base address of a region of memory reserved for use as a TPM's Command Response Buffer region.

As described in the overview text, this is a region of memory that has been carved out and reserved, for example by Trusted Firmware-A.  So, there is no "contents" that would make sense here.  What we want is a way to pass the address of the carved out region between firmware components.

source/transfer_list.rst
stuyod01: Agree this is a good idea.

source/transfer_list.rst
sjg20: Then shouldn't this region be set up within the Transfer List by TF-A? Which program is intended to add this entry to the TL?

stuyod01: Yes, exactly.  TF-A would allocate the reserved region and then add the address/size to the transfer list.  This communicates the info to EDK2/u-boot/secure-partitions so they know where the CRB region is.

sjg20: So then, how about:

1. TF-A creates a TL
2. TF-A adds the TPM block to it (say it is 4KB long)
3. TF-A passed it on to U-Boot, etc.?

i.e. why does it need to be in another area of memory?

stuyod01: This reserved CRB region will be reserved for use by the TPM implementation over the life of the system, and so it's likely you don't want it coming out of a random Normal world address.

Having it in a separate area gives platform's much more flexibility in when and how they reserve memory.  Some platforms may want to statically define reserved regions at specific addresses (for example: https://github.com/ARM-software/arm-trusted-firmware/blob/master/include/plat/arm/common/arm_def.h).  Others may want to reserve regions in a linker script.  There may be other platform specific approaches.

So, we don't want to force all platforms into one approach, unless there is some compelling reason to do this.

1. TF-A platform code reserves CRB memory by platform specific means
2. TF-A adds the address/size to the TL in the TPM info entry
3. TF-A passes it on to U-boot, etc.

sjg20: The TL is intended to hold the actual data so that should be the default approach.

If you put a secure-world address in a normal-world data structure, what happens? Who could actually consume it?

Reservation of the bloblist memory needs to be handled by common code, with other code adding TEs as needed. So the TF-A platform-specific means can add a TE, rather than adding it in a random address.

At least that should be the default.

sjg20: I sent a PR as I now see that the spec didn't include this language, although it was much-discussed and is documented in the U-Boot bloblist implementation.

jmarinho: There are some use-cases where allowing pointers may make sense.

I see 2 compeling ones:
1) some data-structure that can be appended to during boot (e.g. the TPM log). The entity that appends is unaware of the TL, or we don't want to handle potential resizes.
2) Some particular MMIO region could be at a particular fixed address (for some platforms/SoCs the CRB can be hardcoded to some address).

I agree that, in general, TL entries should contain the data.
For the TPM log (and TPM CRB) I wonder if we should open an exception -- as it may be infeasible (for some platforms) to have them within the TL entry.

stuyod01: > If you put a secure-world address in a normal-world data structure, what happens? Who could actually consume it?

A secure world address can not be accessed by normal world, so there would never be a reason to pass an address like that to the normal world.

sjg20: Re the use cases:
1. We can allocate the TPM log to be large enough for initial use.  It is possible to expand a TE.
2. You want to put the TPM log in in a memory-mapped I/O device? What type of device would that be?

I am OK with an exception where needed...but then I think we should have two TEs, the normal case (containing the actual TPM log) and the exception one (for when that is not possible and has to be in strange memory). The reason for needing an exception should not be, for example, "our project has platform-specific code which we don't control and wants to put it whereever it likes".

stuyod01: One distinction we should make is differentiating between passing "data" and a reference to a "resource".

In the case of the proposed TPM Info TE there are 2 separate items that are currently specified by address/size:
1.  TPM event log.  This is "data" that is populated by TF-A measured boot and passed to next boot stage.
2. CRB address/size.   This is a "resource".  There is no data in it.  It's a reference to a reserved region allocated in a platform specific way and we need to pass the address/size to the next stage of boot.  I can't think of any scenario where it would make sense to allocate this in the TL itself.  It's just an empty buffer.

Right now the TPM info combines both items in one TPM info TE.  I would be ok splitting this into 2 distinct TEs, one for the TPM event log and one for the CRB address/size.

sjg20: OK, so perhaps we can agree that the event log should be embedded in a TE, so we should split into two TEs.

For the CRB, that seems OK...but why can't each program use their own CRB address/size? What is special about that address/size? E.g. I might want to use a local var as the buffer, or allocate it with malloc() ? Sorry I don't know much about CRB , but am happy to read a page or two if you can post a link.

stuyod01: There is normally 1 TPM in a system and the CRB is the interface that software will use to send commands to the TPM and receive responses.  So, for Linux the address of the CRB is advertised in an ACPI table.  So software (e.g. Linux) can't pick where the CRB is, it is advertised via ACPI along with other info about the TPM.

If the CRB is in normal DRAM, the most convenient thing to do will be for TF-A to reserve/carve-out a chunk of memory early and then communicate the address/size down to other firmware components.

Not sure there is a good high level summary.
- Linux CRB driver is here: https://github.com/torvalds/linux/blob/master/drivers/char/tpm/tpm_crb.c
- TCG Mobile CRB spec is here: https://trustedcomputinggroup.org/resource/tpm-2-0-mobile-command-response-buffer-interface-specification/.  But, it's more about what the fields in the CRB are and less about how it integrates into a system and who allocates the memory.  There is an intro section that may help.

sjg20: OK, thanks...so are you saying that if the TPM were on PCI we would just use the BARs to find the CRB, but because we are not using PCI, we need to use ACPI (or presumably DT on non-server ARM devices) to specify the address?

It seems from the docs there are two cases:
- TPM chip with MMIO at a particular address
- software TPM with a mailbox at a particular address

Is the address set up using some hardware registers somewhere, like a aperture that you need to select? Can you point to the ACPI and DT bindings?

stuyod01: > OK, thanks...so are you saying that if the TPM were on PCI we would just use the BARs to find the CRB, but because we are not using PCI, we need to use ACPI (or presumably DT on non-server ARM devices) to specify the address?
Yes, the address needs to be advertised by ACPI/DT.

>  It seems from the docs there are two cases:
> TPM chip with MMIO at a particular address
> software TPM with a mailbox at a particular address

For a TPM chip accessed through hardware registers, ACPI/DT will describe it like any other device.  It is the "software TPM" case where a DRAM memory carve out is needed for the CRB.

> Is the address set up using some hardware registers somewhere, like a aperture that you need to select?
No.  For a TPM chip on SPI or I2C, it is addressed like any other SPI/I2C device.  There is no setting of the address.  And the hardware addressing info has to be reflected in ACPI/DT.

For the case of a software TPM in TrustZone, the CRB is just a shared memory communication buffer.  The software TPM learns about the address through a mechanism like the transfer list.  TPM clients like u-boot/EDK2/Linux would learn about the location of CRB through the transfer list, ACPI, or DT.

The ACPI binding is here: https://trustedcomputinggroup.org/resource/tcg-acpi-specification/
I am not aware of a DTB binding for TPM CRB yet.  But, we will need one in the case of a TrustZone-based TPM with a client that uses DT.

sjg20: OK, so I believe you need a TE for SW_TPM_CRB which just has the address and size of this region.

For the buffer, there should be a separate TE the actually contains the buffer.

sjg20:
What is the status of this PR, please?

sjg20:
I don't know what is going with this...perhaps it is just github's terrible review interview, but I cannot see if changes were made, etc. Please can you respond to the threads above?

tusharkhandelwal2:
if you click on Compare link next to the new update message, you should be able to see the delta.

tusharkhandelwal2:
Sorry I am not sure if i have messed up something, I am using github for the first time. But I am following the same process as earlier.

sjg20:
> Sorry I am not sure if i have messed up something, I am using github for the first time. But I am following the same process as earlier.

OK...but just above here I can see some comments I made, with no reply. Do you see those?

tusharkhandelwal2:
I can see this only " What is the status of this PR, please?". Your review comments i can see in my mail only which I guess is sent by github.

sjg20:
> I can see this only " What is the status of this PR, please?". Your review comments i can see in my mail only which I guess is sent by github.

OK please send me an email and I will send a screenshot of what I see - I am sjg@chromium.org

Github review is perhaps a hopeless cause

source/transfer_list.rst
tusharkhandelwal2: This is updated in the new patch

tusharkhandelwal2: Do I have to click on "Resolve Conversation" one i have made the required change?

source/transfer_list.rst
```diff
@@ -657,5 +659,101 @@ such that the last ACPI table in this entry ends at offset
...
+       updated. A component with access to the TPM would replay the event log
+       by reading each measurement recorded and extending it into the TPM. Once
+       the measurements are extended into the TPM, then the "need_to_replay"
+       flag must be cleared if the transfer list is passed to additional
+       firmware components, means bootloader needs to replay the EventLog.
+       Default value is "0".    
+   
+   * - event_log
+     - size of event log.
+     - 0xC
```
sjg20, line 712: 0xc (lower case)

Please fix below too

source/transfer_list.rst
sjg20: I don't know, honestly. Github reviewing is just bizarre to me.

Someone above suggested "The physical base address of a region of memory reserved for use as a TPM's Command Response Buffer region."

but it doesn't look like you added that?

sjg20 (changes requested):

source/transfer_list.rst
tusharkhandelwal2: sure.

sjg20 (dismissed):

source/transfer_list.rst
```diff
@@ -657,5 +659,102 @@ such that the last ACPI table in this entry ends at offset
...
+     - |hdr_size_desc|
+
+   * - data_size
+     - 0x4
+     - 0x4
+     - The size of the event log in bytes + sizeof(flags) i.e. 0x4.
+
+   * - flags
+     - 0x4
+     - 0x8
```
danh-arm, line 697: This fixed offset is not robust to changes in the TE header size. It should be hdr_size like other TEs.

source/transfer_list.rst
```diff
@@ -657,5 +659,102 @@ such that the last ACPI table in this entry ends at offset
...
+       next component in the boot chain is aware that the TPM has not been
+       updated. A component with access to the TPM would replay the event log
+       by reading each measurement recorded and extending it into the TPM. Once
+       the measurements are extended into the TPM, then the "need_to_replay"
+       flag must be cleared if the transfer list is passed to additional
+       firmware components, means bootloader needs to replay the EventLog.
+       Default value is "0".    
+   
+   * - event_log
+     - size of event log.
```
danh-arm, line 711: Would be more consistent with other TEs to say data_size - 0x4

source/transfer_list.rst
```diff
@@ -657,5 +659,102 @@ such that the last ACPI table in this entry ends at offset
...
+     - 0xc
```
danh-arm, line 712: This fixed offset is not robust to changes in the TE header size. It should be hdr_size + 0x4

source/transfer_list.rst
```diff
@@ -657,5 +659,102 @@ such that the last ACPI table in this entry ends at offset
...
+     - |hdr_size_desc|
+
+   * - data_size
+     - 0x4
+     - 0x4
+     - This value should be set to **0xC** i.e. sizeof(crb_base_address) + sizeof(crb_size).
+
+   * - crb_base_address
+     - 0x8
+     - 0x8
```
danh-arm, line 749: This fixed offset is not robust to changes in the TE header size. It should be hdr_size.

source/transfer_list.rst
```diff
@@ -657,5 +659,102 @@ such that the last ACPI table in this entry ends at offset
...
+     - The physical base address of a region of memory reserved for use as a
+       TPM's Command Response Buffer region.
+
+   * - crb_size 
+     - 0x4
+     - 0x10
```
danh-arm, line 755: Should be hdr_size + 0x8

source/transfer_list.rst
```diff
@@ -657,5 +659,102 @@ such that the last ACPI table in this entry ends at offset
...
+.. _tpm_crb_base_entry:
+
+TPM CRB base address table entry layout (XFERLIST_TPM_CRB_BASE)
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+The CRB info defines the address of a region of memory that has been carved out
+and reserved for use as a TPM Command Response Buffer interface.
+
+.. _tab_tpm_crb_base:
+.. list-table:: TPM CRB base type layout
+   :widths: 2 2 2 8
```
danh-arm, line 725: The first column needs to be wider to accommodate crb_base_address, which is running into the next column in the rendered version.

danh-arm (changes requested):

source/transfer_list.rst
```diff
@@ -657,5 +659,102 @@ such that the last ACPI table in this entry ends at offset
...
+       measurements to be extended into a TPM and add them to the TPM event
+       log, but those components are unable access the TPM themselves. In this
+       case, the component should set the "need_to_replay" flag so that the
+       next component in the boot chain is aware that the TPM has not been
+       updated. A component with access to the TPM would replay the event log
+       by reading each measurement recorded and extending it into the TPM. Once
+       the measurements are extended into the TPM, then the "need_to_replay"
+       flag must be cleared if the transfer list is passed to additional
+       firmware components, means bootloader needs to replay the EventLog.
+       Default value is "0".
```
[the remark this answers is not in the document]
danh-arm: Sorry, the PR that adds flags to the TL header (https://github.com/FirmwareHandoff/firmware_handoff/pull/17) is not yet merged. But still it's worth aligning to the language there.

source/transfer_list.rst
```diff
@@ -657,5 +659,102 @@ such that the last ACPI table in this entry ends at offset
...
      - One or more ACPI tables.
 
+
+.. _tpm_evlog_entry:
+
+TPM event log table entry layout (XFERLIST_EVLOG)
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+This entry type holds TPM-related information for a platform. The TPM event log
+info is a region containing a TPM event log as defined by TCG EFI Protocol
+Specification (https://trustedcomputinggroup.org/resource/tcg-efi-protocol-specification/).
```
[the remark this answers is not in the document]
danh-arm: See the references.rst file. You can add the TCG spec reference there then use a link here, e.g. [TCG EFI]_

source/transfer_list.rst
```diff
@@ -657,5 +659,103 @@ such that the last ACPI table in this entry ends at offset
...
+
+   * - hdr_size
+     - 0x1
+     - 0x3
+     - |hdr_size_desc|
+
+   * - data_size
+     - 0x4
+     - 0x4
+     - This value should be set to **0xC** i.e. sizeof(crb_base_address) + sizeof(crb_size).
```
sjg20, line 746: 0xc

sjg20 (approved):
nit: please use lower-case hex

Otherwise this looks fine

sjg20 (approved):

source/transfer_list.rst
```diff
@@ -657,5 +659,103 @@ such that the last ACPI table in this entry ends at offset
...
+     - 0x4
+     - The size of the event log in bytes + sizeof(flags) i.e. 0x4.
+
+   * - flags
+     - 0x4
+     - hdr_size
+     - flags are intended to signal properties of this TE. Bit 0 is
+       need_to_replay flag. Some firmware components may compute measurements
+       to be extended into a TPM and add them to the TPM event log, but those
+       components are unable access the TPM themselves. In this case, the
```
manish-pandey-arm, line 701: nit: unable *to* access

manish-pandey-arm (dismissed):

source/transfer_list.rst
```diff
@@ -657,5 +659,103 @@ such that the last ACPI table in this entry ends at offset
...
+   * - crb_base_address
+     - 0x8
+     - hdr_size
+     - The physical base address of a region of memory reserved for use as a
+       TPM's Command Response Buffer region.
+
+   * - crb_size 
+     - 0x4
+     - hdr_size + 0x8
+     - Size of CRB. If crb_size is "0" that means crb_base_address is not valid.
```
sjg20, line 757: If the address is not valid, then this TE should not be provided, right? Or is it useful to describe the size but not where it is? I think this needs some more explanation

sjg20 (changes requested):

source/transfer_list.rst
stuyod01: We should delete "If crb_size is "0" that means crb_base_address is not valid."

That is a vestige of when this was all combined into one transfer list.  If there is no region reserved for the CRB, then the TE will not be present at all.

sjg20 (dismissed):

manish-pandey-arm (approved):

manish-pandey-arm:
It looks like there are some change requests from @danh-arm which is not done yet. But all the comments from him has already been resolved. Should it be overridden and this being merged?

manish-pandey-arm (approved):

manish-pandey-arm:
Shall we merge this change?

source/transfer_list.rst
```diff
@@ -657,5 +659,103 @@ such that the last ACPI table in this entry ends at offset
...
+
+   * - flags
+     - 0x4
+     - hdr_size
+     - flags are intended to signal properties of this TE. Bit 0 is
+       need_to_replay flag. Some firmware components may compute measurements
+       to be extended into a TPM and add them to the TPM event log, but those
+       components are unable  to access the TPM themselves. In this case, the
+       component should set the "need_to_replay" flag so that the next
+       component in the boot chain is aware that the TPM has not been updated.
```
apalos, line 703: nit, but we could replace 'that the TPM has not been updated.' -> 'the PCRs have not been extended'

source/transfer_list.rst
```diff
@@ -657,5 +659,103 @@ such that the last ACPI table in this entry ends at offset
...
+       A component with access to the TPM would replay the event log by reading
+       each measurement recorded and extending it into the TPM. Once the
+       measurements are extended into the TPM, then the "need_to_replay" flag
+       must be cleared if the transfer list is passed to additional firmware
+       components, means bootloader needs to replay the EventLog. Default
```
apalos, line 708: Why not clear it unconditionally?

tusharkhandelwal2: So if the earlier boot stage doesn't set it, it will be clear only. The component which extend the PCR should only clear it.

danh-arm: This sentence doesn't make sense: "Once the measurements are extended into the TPM, then the "need_to_replay" flag must be cleared if the transfer list is passed to additional firmware components, means bootloader needs to replay the EventLog.". It looks like the last clause ("means bootloader needs to replay the EventLog") should be removed.

danh-arm (changes requested):
Just one comment, otherwise looks good.

sjg20 (dismissed):
LGTM

stuyod01 (approved):

tusharkhandelwal2:
i tried updating my fork , resulted in pull request bring closed. what is the right way to update the fork and rebase the patch ?

jmarinho (approved):

danh-arm (approved):
This one has been going on too long. My minor comments can be addressed later if you agree.

Issue #13, opened 2023-07-26
Thoughts about the checksum field


jwerner-chromium:
@danh-arm @jmarinho @sjg20

I was recently thinking about some possible use cases for the transfer list specification and I think I have some concerns about the checksum field, I thought it might be worth discussing these while we're still in the "v0.9 probation period". The checksum may be fine for normal informational records that get added to the TL once and then mostly stay unchanged forever, but not all use cases may necessarily be like that. A classic example would be something like an in-memory firmware log (e.g. the "CBMEM console" we have in coreboot) that constantly writes more lines into the TE data area for every printf() call. It would be pretty inefficient to recalculate the checksum on the fly every time a new line gets logged, and from an architectural point of view it is likely that the subsystem which modifies the buffer (e.g. the logging code in this example) may want to remain agnostic of where exactly the buffer is stored and that this storage backend has specific requirements about extra steps that need to be done whenever something is updated.

Of course a solution could be that we let the checksum lapse for a bit and then only recalculate it at the end of a firmware stage right before we hand off to the next one. But this still adds a sort of unnecessary tax to adding large data buffers to the TL (e.g. a 128K console log like coreboot allocates by default, most of which is not actually used on most boots and thus just full of uninitialized data that would pointlessly need to get checksummed), particularly because our byte-at-a-time checksumming algorithm is very simple to implement but not really the fastest kind of memory checksum you can have on a 64-bit machine. There might also be a concern about different writers at different exception levels messing around in the same buffer at the same time, which may be okay for that specific buffer but maybe not for the TL metadata structures (e.g. on our Arm Chromebook systems both BL31 and BL33 are technically sharing the same CBMEM console buffer without concurrent access arbitration, which we accept because in practice having both try to log something at the same time is very unlikely and if it does happen the worst it's gonna do is a bit of console output corruption).

Does the checksum really give us enough real benefit to be worth the potential cost of making the TL unsuitable for certain applications? Or did we just include it because of some vague idea that it's "best practice" to have checksums for persistent data structures, without any real practical benefit for our particular design? I mean, the TL is generally just passed around in memory within a single boot cycle, and memory normally doesn't get random bit errors out of nowhere. What kind of corruption are we actually worried about that we want the checksum to guard against?

If we do believe that the checksum is important for certain use cases, should we maybe consider offering both options (e.g. differentiated by different magic numbers) to allow applications to choose whether maintaining a checksum is worth the cost for them?

[...]

PR #17, opened 2023-08-31
Checksum

sjg20:
Here is my attempt to add a checksum bit. There is a bit of clean-up too.

I am not sure how to make the table headings show bold. I can make them into headings, but that changes the font to san-serif, which looks odd. So I left that alone.

### DIFF
--- a/source/register_conventions.rst
+++ b/source/register_conventions.rst
@@ -55,7 +55,7 @@ absent from the transfer list.
    +--------------+-------------------------------------------------------------+
    | X1           | X1 is divided into the following fields:                    |
    |              |                                                             |
-   |              | - X1[23:0]: set to the TL signature (0x6e_d0ff)             |
+   |              | - X1[23:0]: set to the TL signature (0x40_b10b)             |
    |              | - X1[31:24]: version of the register convention used. Set to|
    |              |   1 for the AArch64 convention specified in this document.  |
    |              | - X1[63:32]: reserved, must be zero.                        |
@@ -101,7 +101,7 @@ absent from the transfer list.
    +--------------+-------------------------------------------------------------+
    | R1           | R1 is divided into the following fields:                    |
    |              |                                                             |
-   |              | - R1[23:0]: set to the TL signature (0x6e_d0ff)             |
+   |              | - R1[23:0]: set to the TL signature (0x40_b10b)             |
    |              | - X1[31:24]: version of the register convention used. Set to|
    |              |   1 for the AArch32 convention specified in this document.  |
    |              |                                                             |
--- a/source/transfer_list.rst
+++ b/source/transfer_list.rst
@@ -48,7 +48,8 @@ Transfer list header
 
 A TL must begin with a TL header. The layout of the TL header is shown in
 :numref:`tab_tl_header`.  The presence of a TL header can be verified by
-inspecting the signature field which must contain the 0x6e_d0ff value.  The
+inspecting the signature field which must contain the 0x40_b10b value.  The
+
 version field determines the contents of the handoff start header. The version
 will only be changed by an update to this specification when new TL header or
 TE header fields are defined (i.e. not when allocating new tag IDs), and all
@@ -66,7 +67,7 @@ changes will be backwards-compatible to older readers.
    * - signature
      - 0x4
      - 0x0
-     - The value of signature must be `0x6e_d0ff`.
+     - The value of signature must be `0x40_b10b`.
 
    * - checksum
      - 0x1
@@ -198,7 +199,7 @@ Inputs:
 
 - `tl_base_addr`: Base address of the existing TL.
 
-#. Compare `tl.signature` (`tl_base_addr + 0x0`) to `0x6e_d0ff`. On a mismatch,
+#. Compare `tl.signature` (`tl_base_addr + 0x0`) to `0x40_b10b`. On a mismatch,
    abort (this is not a valid TL).
 
 #. Compare `tl.version` (`tl_base_addr + 0x5`) to the expected version
@@ -353,7 +354,7 @@ Inputs:
 
 #. Check that `available_size` is larger than `0x18` (the assumed `tl.hdr_size`), otherwise abort.
 
-#. Set `tl.signature` (`tl_base_addr + 0x0`) to `0x6e_d0ff`.
+#. Set `tl.signature` (`tl_base_addr + 0x0`) to `0x40_b10b`.
 
 #. Set `tl.checksum` (`tl_base_addr + 0x4`) to `0x0` (for now).
 
--- a/source/transfer_list.rst
+++ b/source/transfer_list.rst
@@ -484,8 +484,8 @@ space up to the next TE).
      - 0x4
      - 0x4
      - The size of the void space in bytes. May be 0. For XFERLIST_VOID,
-     data_size *MUST* be a multiple of 8 (i.e. there must be no space left to
-     inter-TE padding after this TE).
+       data_size *MUST* be a multiple of 8 (i.e. there must be no space left to
+       inter-TE padding after this TE).
 
    * - void_data
      - data_size
--- a/source/transfer_list.rst
+++ b/source/transfer_list.rst
@@ -72,7 +72,7 @@ changes will be backwards-compatible to older readers.
    * - checksum
      - 0x1
      - 0x4
-     - If enabled by the flags, the checksum is used to provide basic protection against something overwriting the TL in memory. The checksum is set to a value such that the sum over every byte in the {`tl_base_pa`, …, `tl_base_pa + size - 1`} address range, modulo 256, is equal to 0. For the purposes of this calculation, the value of this checksum field in the TL header must be assumed as 0. Note that the checksum includes the TL header, all TEs and the inter-TE padding, but not the range reserved for future TE additions up to max_size. The values of inter-TE padding bytes are not defined by this specification and may be uninitialized memory. (This means that multiple TLs with exactly the same size and contents may still have different checksum values.). If checksums are not used, this must be 0.
+     - If enabled by the flags, the checksum is used to provide basic protection against something overwriting the TL in memory. The checksum is set to a value such that the xor over every byte in the {`tl_base_pa`, …, `tl_base_pa + size - 1`} address range, is equal to 0. For the purposes of this calculation, the value of this checksum field in the TL header must be assumed as 0. Note that the checksum includes the TL header, all TEs and the inter-TE padding, but not the range reserved for future TE additions up to max_size. The values of inter-TE padding bytes are not defined by this specification and may be uninitialized memory. (This means that multiple TLs with exactly the same size and contents may still have different checksum values.). If checksums are not used, this must be 0.
 
    * - version
      - 0x1
@@ -216,9 +216,8 @@ Inputs:
    equal to the size of the total area reserved for the TL (if known). If not,
    abort (TL is corrupted).
 
-#. *(optional)* If `has_checksum`, check that the sum of `tl.size` bytes
-   starting at `tl_base_addr` minus `tl.checksum` is equal to `tl.checksum`. If
-   not, abort (TL is corrupted).
+#. *(optional)* If `has_checksum`, check that the xor of `tl.size` bytes
+   starting at `tl_base_addr` is 0x0. If not, abort (TL is corrupted).
 
 Reading a TL
 ^^^^^^^^^^^^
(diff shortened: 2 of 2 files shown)


source/transfer_list.rst
```diff
@@ -357,8 +371,12 @@ Inputs:
 
 #. Set `tl.max_size` (`tl_base_addr + 0xc`) to `available_size`.
 
-#. Calculate the checksum as the sum of all bytes from `tl_base_addr` to
-   `tl_base_addr + tl.hdr_size`, and write the result to `tl.checksum`.
+#. Set `tl.flags` (`tl_base_addr + 0x10`) to `1` if checksums are to be used,
```
danh-arm, line 374: minor: Putting the "If" clause first is more natural for engineers!

danh-arm (changes requested):
Also, we be good to address all points in https://github.com/FirmwareHandoff/firmware_handoff/issues/13:

- Changing checksum algorithm to XOR.
- Changing signature to not use 0x00 or 0xFF in its bytes (and perhaps choose a more memorable value).

source/transfer_list.rst
```diff
@@ -54,6 +54,10 @@ will only be changed by an update to this specification when new TL header or
 TE header fields are defined (i.e. not when allocating new tag IDs), and all
 changes will be backwards-compatible to older readers.
 
+A checksum field is used to provide basic protection against something
+overwriting the TL in memory. It is optional, controlled by a bit in the flags
+word as below.
+
```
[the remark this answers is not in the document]
sjg20: done

source/transfer_list.rst
```diff
@@ -98,6 +98,31 @@ changes will be backwards-compatible to older readers.
...
+     - Flags word. See below for contents.
+
+   * - spare
+     - 0x4
+     - 0x14
+     - Spare word. Must be 0.
+
+
+TL Flags
+--------
```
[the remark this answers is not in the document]
sjg20: done

source/transfer_list.rst
```diff
@@ -112,14 +116,23 @@ changes will be backwards-compatible to older readers.
 TL Flags
 --------
 
+The TL flags word is intended to signal properties relating to the TL as a
+whole. Future flag values may be added according to the rules of the `version`
+field.
+
```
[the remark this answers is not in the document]
sjg20: done

source/transfer_list.rst
```diff
@@ -98,6 +98,31 @@ changes will be backwards-compatible to older readers.
...
 
+   * - flags
+     - 0x4
+     - 0x10
+     - Flags word. See below for contents.
+
+   * - spare
+     - 0x4
+     - 0x14
+     - Spare word. Must be 0.
```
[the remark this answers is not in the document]
sjg20: Hmmm it is going to be hard for me to correct all these things in the context of this change, as each will require a conversation. Can I suggest you do a PR to tidy up the language in the existing spec?

source/transfer_list.rst
```diff
@@ -98,6 +98,31 @@ changes will be backwards-compatible to older readers.
...
+.. list-table:: Flags
+   :widths: 2 2 8
+
+   * - Bit
+     - Name
+     - Description
+
+   * - 31:0
+     - unused
+     - Reserved for future use. Must be 0.
```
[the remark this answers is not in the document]
sjg20: done

source/transfer_list.rst
sjg20: done

sjg20:
> Also, we be good to address all points in #13:
> 
> * Changing checksum algorithm to XOR.
> * Changing signature to not use 0x00 or 0xFF in its bytes (and perhaps choose a more memorable value).

OK I added two more commits

source/transfer_list.rst
```diff
@@ -98,6 +98,35 @@ changes will be backwards-compatible to older readers.
...
+     - 0x10
+     - Flags word. See below for contents.
+
+   * - reserved
+     - 0x4
+     - 0x14
+     - Reserved word. Must be set to 0 or ignored.
+
+
+TL Flags
```
jmarinho, line 114: Do we have the need for the flags field today?
having whole-TL flags is valuable, and we'll certainly need this at some point.
Until we have a concrete need for the flag, is it best to leave the field as "reserved"? -- The reserved field would effectively be 8-bytes wide.

If we do leave it entirely as reserved, we could comment the src saying that we intended to use part of the field for flags.

source/register_conventions.rst
```diff
@@ -55,7 +55,7 @@ absent from the transfer list.
    +--------------+-------------------------------------------------------------+
    | X1           | X1 is divided into the following fields:                    |
    |              |                                                             |
-   |              | - X1[23:0]: set to the TL signature (0x6e_d0ff)             |
+   |              | - X1[23:0]: set to the TL signature (0x40_b10b)             |
```
jmarinho, line 58: The reasoning makes sense to me. And the proposed signature does too.
+1

jmarinho (dismissed):
FWIW, I'm in favour of us adopting a 'xor' checksum as that provides more implementation flexibility versus the 'sum' scheme we had before, and still meet the requirements.

sjg20:
> Do we have the need for the flags field today?
> having whole-TL flags is valuable, and we'll certainly need this at some point.
> Until we have a concrete need for the flag, is it best to leave the field as "reserved"? -- The reserved field would > effectively be 8-bytes wide.

> If we do leave it entirely as reserved, we could comment the src saying that we intended to use part of the field for flags.

We need one bit of the flag...are you suggesting that we reduce the flags to only 8 bits, or something else?

@jmarinho

jmarinho:
> We need one bit of the flag...are you suggesting that we reduce the flags to only 8 bits, or something else?
@sjg20 
I was too quick to add the comment.
Only afterwards did I notice that the flags word was being put to use in a following commit :)

source/transfer_list.rst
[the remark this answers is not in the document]
sjg20: I changed the language to reserved, so at least for this PR, this is resolved

source/transfer_list.rst
```diff
@@ -66,12 +67,13 @@ changes will be backwards-compatible to older readers.
    * - signature
      - 0x4
      - 0x0
-     - The value of signature must be `0x6e_d0ff`.
+     - The value of signature must be `0x40_b10b`.
```
jwerner-chromium, line 70: Wasn't part of the reason for changing the signature that it makes more sense to use all 4 bytes? Now we're again leaving the top byte 0 which seems like a waste? I'm aware that this is also used in the 3-byte field of the register convention, but I still don't think that's a good reason to leave it 0x00 here. Or if we do decide that 3 bytes are enough we might as well use the remaining byte for something useful instead, like the flags (or maybe the version field should be moved here which would make it very similar to how the top byte is used in the register convention).

source/transfer_list.rst
```diff
@@ -98,6 +100,44 @@ changes will be backwards-compatible to older readers.
...
+
+TL Flags
+^^^^^^^^
+
+The TL flags word is intended to signal properties relating to the TL as a
+whole. Future flag values may be added according to the rules of the `version`
+field.
+
+The TL flags word is intended to signal properties relating to the TL as a
+whole. Future flag values may be added according to the rules of the `version`
```
jwerner-chromium, line 122: This text seems to appear twice?

source/transfer_list.rst
jwerner-chromium: Sorry, I don't understand what you mean. Didn't we agree on the solution that increases the TL header size by 8 bytes now? Isn't that what this patch is doing?

As far as I can tell the signature field here still has 4 bytes, it's just that one of them is (still) 0 and I think it could be something more distinctive.

sjg20:
> Also, we be good to address all points in #13:
> 
> * Changing checksum algorithm to XOR.
> * Changing signature to not use 0x00 or 0xFF in its bytes (and perhaps choose a more memorable value).

OK, done

sjg20:
Are there any more comments on this one?

danh-arm (approved):
LGTM. I'd like Julius's +1 though.

source/transfer_list.rst
```diff
@@ -71,7 +71,8 @@ changes will be backwards-compatible to older readers.
    * - checksum
      - 0x1
      - 0x4
-     - The checksum is used to ensure the data contained within the list is intact. The checksum is set to a value such that the sum over every byte in the {`tl_base_pa`, …, `tl_base_pa + size - 1`} address range, modulo 256, is equal to 0. For the purposes of this calculation, the value of this checksum field in the TL header must be assumed as 0. Note that the checksum includes the TL header, all TEs and the inter-TE padding, but not the range reserved for future TE additions up to max_size. The values of inter-TE padding bytes are not defined by this specification and may be uninitialized memory. (This means that multiple TLs with exactly the same size and contents may still have different checksum values.)
+     - If enabled by the flags, the checksum is used to provide basic protection
```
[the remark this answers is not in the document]
sjg20: Not really, it's just that I don't like 700-character lines, so I split it.

In general I think we should stick to 80cols for easier editing.

sjg20:
> LGTM. I'd like Julius's +1 though.

OK...but it looks like there are a few other PRs that this will affect, so we are a bit stuck

danh-arm:
@jwerner-chromium Could you review this since it addresses the issues you originally raised in https://github.com/FirmwareHandoff/firmware_handoff/issues/13?

sjg20:
we have a few other PRs which are racing this one...what shall we do?

source/transfer_list.rst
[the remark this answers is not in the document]
jwerner-chromium: It's not really split to 80cols consistently now though, is it? It looks (or is that a GitHub display issue?) like you just split the first line but then the second line still contains everything on one line.

I'm fine with splitting at 80cols but then it should be consistent (at least within the same block of text).

source/transfer_list.rst
```diff
@@ -181,8 +221,9 @@ Inputs:
    equal to the size of the total area reserved for the TL (if known). If not,
    abort (TL is corrupted).
 
-#. *(optional)* Check that the sum of `tl.size` bytes starting at `tl_base_addr`
-   minus `tl.checksum` is equal to `tl.checksum`. If not, abort (TL is corrupted).
+#. *(optional)* If `has_checksum`, check that the xor of `tl.size` bytes
+   starting at `tl_base_addr` minus `tl.checksum` is equal to `tl.checksum`. If
```
jwerner-chromium, line 225: I think the "minus `tl.checksum`" isn't quite correct now anymore, that would also be an XOR, but if you write it like that I'm worried it might be a bit hard to read.

I think with XOR we should be able to just use the simpler alternative "check that the XOR of `tl.size` bytes starting at `tl_base_addr` is `0x0`" (because the value already in the checksum field should exactly cancel out the XOR of everything else excluding the checksum field), right?

jwerner-chromium (approved):
Sorry, was a bit busy. This looks good to me except for two more nits.

danh-arm:
@sjg20 - I'll give you a day or so to decide whether to address Julius's minor comments, then I'll merge (ahead of the other PRs).

source/transfer_list.rst
jmarinho: The text duplication still apears in the latest commit. I'll "unresolve" the comment so we don't lose track.

jmarinho (dismissed):
LGTM, other than a minor text dupplication that should be fixed before merge.

jmarinho:
Hi @sjg20 ,
Just confirmed, in the artifact generated from the current commit, the following paragraph is duplicated:
"The TL flags word is intended to signal properties relating to the TL as a
whole. Future flag values may be added according to the rules of the `version`" is duplicated.

I think this may be fixed easily, and so was wondering if that can be done before we merge this PR!

image

Otherwise since all patch reviewers accepted it the PR can be merged.

source/transfer_list.rst
sjg20: It might be a rebase issue. I'm not sure, but it seems gone for me

source/transfer_list.rst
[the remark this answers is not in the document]
sjg20: OK, I was trying to minimise changes. I'll just give up and leave it long

source/transfer_list.rst
sjg20: OK done

sjg20:
OK I updated this. I don't see the duplication, but if you still do, let me know

jmarinho (approved):
Thank you Simon

apalos (approved):

PR #18, opened 2023-09-19
Advise that pointers should not be included

sjg20:
This language did not make it into the initial spec. Add some guidance against pointers and addresses.

### DIFF
--- a/source/transfer_list.rst
+++ b/source/transfer_list.rst
@@ -181,6 +181,29 @@ The TE header is defined in :numref:`tab_te_header`.
      - The exact size of the data content in bytes, not including inter-TE padding. May be 0.
 
 
+TL Contents
+-----------
+
+The TL must not hold pointers or addresses within its entries, which refer to
+anything in the TL. These can make it difficult to relocate the TL. TL
+relocation typically happens in later phases of the boot when there is more
+memory available, which is needed for adding larger entries.
+
+The TL may hold pointers or addresses which refer to regions outside the TL, if
+this is necessary. For example, the MMIO address of a device may be included in
+a TE. But in general, pointers and addresses should be avoided. Instead, the
+data structure itself should generally be contained within the TL. This approach
+provides the greatest flexibility for later boot stages to handle memory as they
+wish, since relocating the TL is fairly simple and self-contained, without
+needing to consider relocating other data structures strewn around the memory.
+
+Where pointers or addresses are needed due to some project-specific restriction,
+a separate TE should generally be created for that purpose, rather than mixing
+pointers with other data. Of course there may be exceptions where two pointers
+belong together, or there is a pointer and a size which belong together. In any
+case, the PR should clearly document the need for these pointers.
+
+
 .. _sec_operations:
 
 Standard operations
--- a/source/transfer_list.rst
+++ b/source/transfer_list.rst
@@ -184,6 +184,31 @@ The TE header is defined in :numref:`tab_te_header`.
 TL Contents
 -----------
 
+Tags are expected to have a simple layout (representable by a C structure) and
+each tag should only represent data for a single logical concept. Data for
+multiple distinct concepts should be split across different tags, even if
+they're always expected to appear together on the first platform adding the tag
+(to encourage reusability in different situations). Alternatively, complex data
+may be represented in a different kind of well-established handoff data
+structure (e.g. FDT [DT]_, HOB [PI]_) that is inserted into the TL as a single
+TE. The same tag ID may occur multiple times in the TL to represent multiple
+instances of the same kind of object. Tag layouts (including the meaning of all
+fields) are considered stable after being added to this specification and may
+never be changed in a backwards-incompatible way. If a backwards-incompatible
+change is desired, a new tag ID should be allocated for the new version of the
+layout instead.
+
+Tag layouts may be changed in a backwards-compatible manner by allowing new
+valid values in existing fields (including reserved fields), as long as the
+original layout definition clearly defined how unknown values in those fields
+should be handled, and the rest of the TE would still be considered valid and
+correct for older readers that consider the new values unknown. TE layouts may
+also be expanded by adding new fields at the end, with the same restrictions.
+TEs should not contain explicit version numbers and instead just use the
+`data_size` value to infer how many fields exist. TE layouts which have been
+changed like this must clearly document which fields or valid values were added
+at a later time, and in what order.
+
 The TL must not hold pointers or addresses within its entries, which refer to
 anything in the TL. These can make it difficult to relocate the TL. TL
 relocation typically happens in later phases of the boot when there is more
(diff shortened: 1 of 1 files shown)


source/introduction.rst
```diff
@@ -67,6 +67,13 @@ off the list to a next stage is termed the *Sender*. The firmware stage
 receiving the list is termed the *Receiver*. The *Receiver* can update fields in
 any entry in the list and is allowed to remove entries.
 
+The TL should not hold pointers or addresses within its entries. These can make
```
stuyod01, line 70: I don't understand how **physical addresses** make it difficult to relocate a transfer list.  Nothing would need to change at all on a relocation.

I do understand that if there were offsets or virtual addresses it would be a problem.

jwerner-chromium:
I think this language is too harsh, especially if we take it as an invitation to refuse patches from people who want to add TEs like this to their project. I think it would be fine to add a general recommendation for allocating data inline where possible and listing the benefits of that approach, but the ultimate decision should remain with the contributing project and their specific needs. During a review, I think it should be fine to ask the contributor to consider if inline data would work for them, but if they disagree we should let them submit TEs referencing outside data if they want to. (And justifications like "we have a vendor blob that we can't change that needs this allocated in a specific area" is absolutely a perfectly valid reason to require an external reference TE. Unchangeable vendor blobs are a reality that many projects need to deal with whether we're happy about it or not.)

This comes down to the question of whether we want this specification to be widely accepted or whether we want to play design police that holds submissions hostage for every project using it (which would not be very many in that case). You cannot effectively have both. A solution that seeks to be universal to all sorts of firmware projects will need to accept that some of them may have different requirements, different ways of approaching things or simply limited resources to invest in supporting an alternative solution, and many of them will like need to tie in to legacy solutions that already exist in some specific way and cannot easily be changed. We can use reviews to effect minor design improvements in the TE itself (e.g. changing the size or arrangement of fields, or whether things should be split out into multiple independent TEs) that the submitter could quickly fix with minimal impact to what they're trying to do, but we cannot use them to force submitters into major design changes (such as how and where to allocate persistent data buffers), or we will quickly find that nobody wants to use a new, untested solution that is such a pain to work with and so restrictive in what you can do with it. We should instead allow each project to submit TE definitions for their specific needs, accepting that this may lead to very case-specific TE definitions that cannot be easily reused (such as TEs with pointers to external data structures that are hard to keep track of) and multiple competing TE definitions for the same concept, and then later let them dynamically decide which one to standardize on if there is desire for interoperation in that area.

sjg20:
I don't really agree with this...we have the private area for private things.

The original U-Boot implementation has separate parts for project-specific and vendor-specific areas. We currently only have 'standard' and 'non-standard'. We could allocate space for project-specific things to handle the case you mention here, i.e. something we don't like, don't want to promote as the 'right way', but accept because the project can't change.

As to the vendor blobs, if we have to have to deal with something like that, we could add it in its own TE and note that it is needed due to that blob.

I will update the language to try to capture that.

sjg20:
Where are we with this one?

danh-arm (changes requested):
Language is too restrictive for me

sjg20:
> I agree with Julius that the language is too restrictive here. Pointing to data outside the TL is not really a special project/vendor thing, it's a common use-case.

Can you give some examples?

> The data being pointed to may have a lifetime/scope outside the control of TL senders.

That doesn't actually matter though. If the data moves on beyond the scope of the TL sender, it still exists in memory somewhere. It's just that we don't know about the data structure around it.

> I would be OK with some words to say that data should be fully contained within the TL where possible but that's about as far as I'd go.

That sounds reasonable to me.

source/introduction.rst
danh-arm: Stuart's comment (which I agree with) hasn't been addressed in this version. If the TL points to memory outside the TL (which is legitimate) then the address is unaffected by the relocation. If you're concerned about pointers to data within the TL (or at least to the same memory region to be relocated), then this absolutely should be banned. In that case, this specific case should be stated, and the "should" can be changed to "must".

danh-arm (changes requested):
> Can you give some examples?

Pointers to MMIO devices (where DT or similar is not desired), pointers to variable sized buffers that need to remain in memory beyond the lifetime of the TL.

>> The data being pointed to may have a lifetime/scope outside the control of TL senders.

> That doesn't actually matter though. If the data moves on beyond the scope of the TL sender, it still exists in memory somewhere. It's just that we don't know about the data structure around it.

It's more likely the data won't move. The definition of the data structure being pointed to should be defined somewhere. We could make it a requirement for the TE definition to refer to this structure definition when such pointers are used (I don't mind either way).

sjg20:
See what you think of this version

sjg20:
OK this is ready for another look

sjg20:
PTAL thanks

stuyod01 (approved):
Looks good to me.

jwerner-chromium:
I think this is a lot better but I'm not quite sure what the last paragraph means. What exactly does "a separate TE must be created" mean... do you mean there needs to be a separate TE for each pointer? What about cases where two pointers belong together, or where a pointer should be accompanied by a small portion of inline metadata? I don't see why those cases should be forbidden.

I think the important part here is that we need to allow leniency for projects that want to port existing data structures to the TL format, even if they don't 100% align with the best practices for new entries. If e.g. the coreboot project were to decide to port its existing coreboot table infrastructure to the TL format, they should be allowed to just dump all their 60+ tag structures in as-is, without having each of them carefully scrutinized to conform to the rules. Things could still be deprecated and replaced piece-by-piece later, but if we start out from the get-go with "you need to do this huge amount of reworks to even start using this spec for your project", nobody is going to adopt it.

sjg20:
> I think this is a lot better but I'm not quite sure what the last paragraph means. What exactly does "a separate TE must be created" mean... do you mean there needs to be a separate TE for each pointer? What about cases where two pointers belong together, or where a pointer should be accompanied by a small portion of inline metadata? I don't see why those cases should be forbidden.

Since pointers may turn into inline data it makes sense to keep them separate. You are referring to corner cases which would come up in review. If there are two related pointers, then it would make sense to put them in the same TE. Same with metadata.

> 
> I think the important part here is that we need to allow leniency for projects that want to port existing data structures to the TL format, even if they don't 100% align with the best practices for new entries. If e.g. the coreboot project were to decide to port its existing coreboot table infrastructure to the TL format, they should be allowed to just dump all their 60+ tag structures in as-is, without having each of them carefully scrutinized to conform to the rules. Things could still be deprecated and replaced piece-by-piece later, but if we start out from the get-go with "you need to do this huge amount of reworks to even start using this spec for your project", nobody is going to adopt it.

From what I know of that format, it would suit this table OK and most of the elements are simple C structures. But we need to review additions for coreboot, just like any other project. It would not be compatible with what coreboot currently uses, in any case, although I am sure that reusing structs would reduce code churn. But would you ever be able to turn off the old format?

sjg20:
> > I think this is a lot better but I'm not quite sure what the last paragraph means. What exactly does "a separate TE must be created" mean... do you mean there needs to be a separate TE for each pointer? What about cases where two pointers belong together, or where a pointer should be accompanied by a small portion of inline metadata? I don't see why those cases should be forbidden.
> 
> Since pointers may turn into inline data it makes sense to keep them separate. You are referring to corner cases which would come up in review. If there are two related pointers, then it would make sense to put them in the same TE. Same with metadata.
> 
> > I think the important part here is that we need to allow leniency for projects that want to port existing data structures to the TL format, even if they don't 100% align with the best practices for new entries. If e.g. the coreboot project were to decide to port its existing coreboot table infrastructure to the TL format, they should be allowed to just dump all their 60+ tag structures in as-is, without having each of them carefully scrutinized to conform to the rules. Things could still be deprecated and replaced piece-by-piece later, but if we start out from the get-go with "you need to do this huge amount of reworks to even start using this spec for your project", nobody is going to adopt it.
> 
> From what I know of that format, it would suit this table OK and most of the elements are simple C structures. But we need to review additions for coreboot, just like any other project. It would not be compatible with what coreboot currently uses, in any case, although I am sure that reusing structs would reduce code churn. But would you ever be able to turn off the old format?

OK I tried to soften it a bit. PTAL

jmarinho (approved):
I quite like the message we converged to!

sjg20:
> Thanks, I think the text sounds fine now. However, I just realized that this is going into the introduction header of the document, while all the other restrictions and explanations on what is and isn't allowed in a TE come later in the "Entry type allocation" section. Those should probably be combined in the same place somehow, either all up here or all down there or by creating a new section somewhere. Could be done in a follow-up PR as well.

Yes I was thinking that too.

I've moved it down and also move the tag-allocation info up a bit.

source/transfer_list.rst
```diff
@@ -181,6 +181,99 @@ The TE header is defined in :numref:`tab_te_header`.
...
+Tag IDs must be allocated in this specification before use. A new tag ID can be
+allocated by submitting a pull request to this repository that adds a
+description of the respective TE data layout to this specification. Tag IDs do
+not have to be allocated in order. Submitters are encouraged to try to group
+tag IDs together in logical clusters at 16 or 256-aligned boundaries (e.g. all
+tags related to a particular chipset or to a particular firmware project could
+use adjacent tag numbers), but there are no predefined ranges and no
+reservations of tag ranges for specific use.
+
+Tags are expected to have a simple layout (representable by a C structure) and
```
jwerner-chromium, line 196: I feel like everything from here...

source/transfer_list.rst
```diff
@@ -181,6 +181,99 @@ The TE header is defined in :numref:`tab_te_header`.
...
+Tag layouts may be changed in a backwards-compatible manner by allowing new
+valid values in existing fields (including reserved fields), as long as the
+original layout definition clearly defined how unknown values in those fields
+should be handled, and the rest of the TE would still be considered valid and
+correct for older readers that consider the new values unknown. TE layouts may
+also be expanded by adding new fields at the end, with the same restrictions.
+TEs should not contain explicit version numbers and instead just use the
+`data_size` value to infer how many fields exist. TE layouts which have been
+changed like this must clearly document which fields or valid values were added
+at a later time, and in what order.
```
jwerner-chromium, line 219: ...to here would fit better under the "Tag Contents" umbrella, if we're introducing that as a separate section. (Also it would probably look better to have Tag Contents before Tag Allocation in that case.)

sjg20:
ping for review, please

stuyod01 (approved):

stuyod01 (approved):

jmarinho (approved):

danh-arm (approved):

danh-arm:
Hi @jwerner-chromium. We'll merge this in the next couple of days unless you request changes.

jwerner-chromium (approved):

PR #19, opened 2023-09-26
add standard TEs for OPTEE

raymo200915:
Add XFERLIST_OPTEE_PAGABLE_PART for OPTEED pagable part address.
Add XFERLIST_SPMC_MANIFEST for SPMC manifest.


source/transfer_list.rst
```diff
@@ -657,5 +660,116 @@ such that the last ACPI table in this entry ends at offset
...
+   * - data_size
+     - 0x4
+     - 0x4
+     - The size of OPTEE pagable part in bytes.
+
+   * - pagable_part
+     - data_size
+     - hdr_size
+     - Pagable part address for OPTEED
+
```
jwerner-chromium, line 699: Do you not want to store the size of the pointed-to memory area anywhere? Is that not relevant for these?

source/transfer_list.rst
```diff
@@ -657,5 +660,116 @@ such that the last ACPI table in this entry ends at offset
...
+   * - data_size
+     - 0x4
+     - 0x4
+     - The size of OPTEED AARCH mode in bytes.
+
+   * - opteed_aarch
+     - data_size
+     - hdr_size
+     - Holds the OPTEED AARCH mode.
```
jwerner-chromium, line 772: You should clarify what exactly is stored here (e.g. is it an enum value? Is it a literal `32` or `64` integer?).

Assuming this is just aarch32 vs aarch64(?) and can thus be represented as one bit, you may want to consider defining it as a bit field so that the other values are available for future expansion.

source/transfer_list.rst
raymo200915: Not sure if I catch what you mean, but if for the size of 'pagable part address', it is stored in 'data_size'.

source/transfer_list.rst
raymo200915: I will prefer to use 4 bytes here if there are no special requirements, as right now OPTEED uses a type of 'uint32_t' to represent aarch mode during setup, in this case, we don't need to interpret the TE data after we parse from the TL.

source/transfer_list.rst
```diff
@@ -657,5 +660,116 @@ such that the last ACPI table in this entry ends at offset
...
+     - |hdr_size_desc|
+
+   * - data_size
+     - 0x4
+     - 0x4
+     - The size of OPTEE pagable part in bytes.
+
+   * - pagable_part
+     - data_size
+     - hdr_size
```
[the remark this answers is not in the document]
raymo200915: hmmm, I guess it is better to align with other existing TEs. I can add "... must be set to 8" in the description.

source/transfer_list.rst
```diff
@@ -657,5 +660,116 @@ such that the last ACPI table in this entry ends at offset
...
+
+   * - Field
+     - Size (bytes)
+     - Offset (bytes)
+     - Description
+
+   * - tag_id
+     - 0x3
+     - 0x0
+     - The tag_id field must be set to **5**.
```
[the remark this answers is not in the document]
raymo200915: Fixed.

source/transfer_list.rst
```diff
@@ -447,9 +447,9 @@ The following entry types are currently defined:
 - single HOB block entry: tag_id = 2 (:numref:`hob_block_entry`).
 - HOB list entry: tag_id = 3 (:numref:`hob_list_entry`).
 - ACPI table aggregate entry: tag_id = 4 (:numref:`acpi_aggr_entry`).
-- OPTEE pagable part entry: tag_id = 5 (:numref:`optee_pagable_part_entry`).
-- TOS FW config entry: tag_id = 6 (:numref:`tos_fw_config_entry`).
-- OPTEED AARCH entry: tag_id = 7 (:numref:`opteed_aarch_entry`).
+- OPTEE pagable part entry: tag_id = 0x100 (:numref:`optee_pagable_part_entry`).
```
sjg20, line 450: I don' t think we need 256 entries for OP-TEE. I suggest using 0x10 or 0x20

source/transfer_list.rst
```diff
@@ -702,8 +702,8 @@ This address is used when OPTEED is the Secure Payload Dispatcher.
 TOS FW config entry layout (XFERLIST_TOS_FW_CONFIG)
 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
 
-This entry type holds the address of SPMC manifest image.
-This address is used when OPTEE represents SPMC at S-EL1 and SPMD is the Secure Payload Dispatcher.
+This entry type holds the SPMC manifest image.
```
sjg20, line 705: What is SPMC ?

source/transfer_list.rst
```diff
@@ -702,8 +702,8 @@ This address is used when OPTEED is the Secure Payload Dispatcher.
...
+It is used when OPTEE represents SPMC at S-EL1 and SPMD is the Secure Payload Dispatcher.
```
sjg20, line 706: What is SPMD?

What is S-EL1 ?

source/transfer_list.rst
```diff
@@ -732,15 +732,15 @@ This address is used when OPTEE represents SPMC at S-EL1 and SPMD is the Secure
...
-     - TOS FW config address for SPMD
+     - TOS FW config base address
 
 .. _opteed_aarch_entry:
 
 OPTEED AARCH entry layout (XFERLIST_OPTEED_AARCH)
 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
 
 This entry type holds the AARCH mode for OPTEED.
-This AARCH mode is used when OPTEED is the Secure Payload Dispatcher.
```
sjg20: What is OPTEED?

In general, if you an acronym, please can you write it out in full at least once?

source/transfer_list.rst
```diff
@@ -657,5 +660,116 @@ such that the last ACPI table in this entry ends at offset
      - hdr_size
      - One or more ACPI tables.
 
+.. _optee_pagable_part_entry:
+
+OPTEE pagable part entry layout (XFERLIST_OPTEE_PAGABLE_PART)
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+This entry type holds the address of OPTEE pagable part.
```
sjg20, line 668: I am struggling to review this as it seems to be split over two commits? Is there a way to see what is going on?

What is a pagable part?

source/transfer_list.rst
```diff
@@ -657,5 +660,116 @@ such that the last ACPI table in this entry ends at offset
...
+This address is used when OPTEED is the Secure Payload Dispatcher.
+
+.. _tab_optee_pagable_part:
+.. list-table:: OPTEE pagable part type layout
```
sjg20, line 672: I am not sure what these terms are. Please can you define them fully?

source/transfer_list.rst
sjg20: But as Julius says, you do need to define the values

sjg20 (changes requested):

source/transfer_list.rst
raymo200915: How about to start from 0x20? Just want to reserve more spaceholders for the "regular" ones.

source/transfer_list.rst
raymo200915: Definitions added.

source/transfer_list.rst
raymo200915: Added the definitions.

source/transfer_list.rst
raymo200915: Added the definitions.

source/transfer_list.rst
raymo200915: Added the definitions.

source/transfer_list.rst
raymo200915: I just did a sqush rebase. Now all changes are in one commit.
Definitions added for 'pagable part'.

source/transfer_list.rst
raymo200915: Added more details explaination.

source/transfer_list.rst
```diff
@@ -657,5 +660,120 @@ such that the last ACPI table in this entry ends at offset
      - hdr_size
      - One or more ACPI tables.
 
+.. _optee_pagable_part_entry:
+
+OP-TEE pagable part entry layout (XFERLIST_OPTEE_PAGABLE_PART)
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+This entry type holds the address of OP-TEE pagable part.
+This address (of type 'uint64_t') is used when OPTEED(OP-TEE Dispatcher)
```
danh-arm, line 669: minor: space after OPTEED

source/transfer_list.rst
```diff
@@ -657,5 +660,120 @@ such that the last ACPI table in this entry ends at offset
...
+     - The size of OP-TEE pagable part in bytes must be set to **8**.
+
+   * - pagable_part
+     - data_size
+     - hdr_size
+     - Pagable part address for OPTEED
+
+.. _tos_fw_config_entry:
+
+TOS FW config entry layout (XFERLIST_TOS_FW_CONFIG)
```
danh-arm, line 704: TOS_FW_CONFIG is a TF-A concept intended to contain all the config data for a TOS. But it seems you're using it here to specifically contain the SPMC manifest, which is a standard FF-A spec structure, not specific to OP-TEE. I think you should create an SPMC manifest TE, which refers to the FF-A spec. This can have a top level tag_id instead of being scoped to be in the OP-TEE space.

source/transfer_list.rst
```diff
@@ -657,5 +660,120 @@ such that the last ACPI table in this entry ends at offset
...
+     - 0x1
+     - 0x3
+     - |hdr_size_desc|
+
+   * - data_size
+     - 0x4
+     - 0x4
+     - The size of OPTEED AARCH mode in bytes must be set to **4**.
+
+   * - opteed_aarch
```
danh-arm, line 773: Why do you need a separate TE for this? Can't this be merged with XFERLIST_OPTEE_PAGABLE_PART and use a more generic name to represent the data that the OPTEED passes to OP-TEE.

source/transfer_list.rst
```diff
@@ -657,5 +660,120 @@ such that the last ACPI table in this entry ends at offset
...
+
+   * - data_size
+     - 0x4
+     - 0x4
+     - The size of TOS FW config in bytes.
+
+   * - tos_fw_config
+     - data_size
+     - hdr_size
+     - TOS FW config base address
```
danh-arm, line 738: Why is this referring to a base address rather then embedding the manifest inline in the TE?

danh-arm (changes requested):

source/transfer_list.rst
raymo200915: Sounds good, but now all OP-TEE specific TEs are starting from 0x20, if we use it for a general purpose to refer to FF-A, I will move it to the general ones staring from 0. Is it good for you?

source/transfer_list.rst
raymo200915: Generally this TE is used in between BL2 and opteed within TF-A scope only, but will not be in OP-TEE, which is not the same case comparing with  `XFERLIST_OPTEE_PAGABLE_PART`.

source/transfer_list.rst
raymo200915: Yes, you are right. This is holding the complete manifest image. I will amend this in the next commit.

source/transfer_list.rst
danh-arm: Sounds good, although we have other in-flight PRs that add new TEs so we might have change the number depending on which is merged first.

source/transfer_list.rst
```diff
@@ -447,6 +447,9 @@ The following entry types are currently defined:
 - single HOB block entry: tag_id = 2 (:numref:`hob_block_entry`).
 - HOB list entry: tag_id = 3 (:numref:`hob_list_entry`).
 - ACPI table aggregate entry: tag_id = 4 (:numref:`acpi_aggr_entry`).
+- OP-TEE pagable part entry: tag_id = 0x20 (:numref:`optee_pagable_part_entry`).
+- TOS FW config entry: tag_id = 0x21 (:numref:`tos_fw_config_entry`).
+- OPTEED AARCH entry: tag_id = 0x22 (:numref:`opteed_aarch_entry`).
```
danh-arm, line 452: I don't think it's scalable to list every single TE here. Perhaps just just link to the OP-TEE section rather than list all OP-TEE TEs.

source/transfer_list.rst
```diff
@@ -657,5 +660,120 @@ such that the last ACPI table in this entry ends at offset
      - hdr_size
      - One or more ACPI tables.
 
+.. _optee_pagable_part_entry:
```
danh-arm, line 663: I think it would be good to start a new sub-section for the OP-TEE TEs.

source/transfer_list.rst
```diff
@@ -657,5 +660,120 @@ such that the last ACPI table in this entry ends at offset
...
+     - The size of TOS FW config in bytes.
+
+   * - tos_fw_config
+     - data_size
+     - hdr_size
+     - TOS FW config base address
+
+.. _opteed_aarch_entry:
+
+OPTEED(OP-TEE Dispatcher) AARCH entry layout (XFERLIST_OPTEED_AARCH)
```
danh-arm, line 742: minor: space after OPTEED here too

danh-arm (changes requested):
Just a few other minor comments

source/transfer_list.rst
raymo200915: Done.

source/transfer_list.rst
raymo200915: Fixed.

source/transfer_list.rst
raymo200915: Done.

source/transfer_list.rst
raymo200915: Done.

source/transfer_list.rst
raymo200915: Fixed.

raymo200915:
All comments have been addressed. Squashed and pushed the commit.

source/transfer_list.rst
```diff
@@ -657,5 +659,121 @@ such that the last ACPI table in this entry ends at offset
...
+
+   * - pagable_part
+     - data_size
+     - hdr_size
+     - Pagable part address for OPTEED
+
+**OPTEED AARCH entry layout (XFERLIST_OPTEED_AARCH)**
+
+This entry type holds the AARCH mode for OPTEED (OP-TEE Dispatcher).
+This AARCH mode (of type 'uint32_t', **0** for AARCH64 and **1** for AARCH32)
```
sjg20, line 746: Is this different from the current CPU mode? I would have thought that you could just check the CPU flags, or even hard-code it since the build system knows the code it is building for.

So please could you add in an explanation for why this is needed?

sjg20 (changes requested):

source/transfer_list.rst
jwerner-chromium: I would recommend to put those definitions down here in the entry for this specific field, not at the top of the TE (for TEs with one field it doesn't make a ton of difference, but since we may also get larger TEs with more fields it would make sense to standardize on keeping everything with the field it belongs to).

source/transfer_list.rst
jwerner-chromium: Sorry, I think there may be a misunderstanding of how the `data_size` field works here? `data_size` always only contains the size of the TE itself (minus `hdr_size`). So if your TE just contains an 8-byte pointer to memory that's somewhere outside of the TL, then the `data_size` field for that TE must be set to 8. But you may still want to store the size of that pointed-to memory area somewhere (unless you really don't need that for your use case here?).

source/transfer_list.rst
jwerner-chromium: Wouldn't it be best to have a separate "Arm" or "FF-A" block of tag IDs for this then? If this specification is truly supposed to be universal (e.g. include x86 PCs), then I wouldn't expect FF-A specific definitions to go into the "general" block. (Especially if we only leave 32 IDs total for that, I think that's a bad idea in general, I would at least leave 256 and start the next block at 0x100...)

source/transfer_list.rst
raymo200915: The reason I am doing in this way is because the current "legacy" handoff from BL2 and opteed is passing the AARCH mode via X0/R0, but not directly reading from SPSR.
If you agree, I can remove this TE from the spec, and then modify the code to allow opteed to get the AARCH mode from SPSR instead of X0/R0 from BL2.

sjg20: > The reason I am doing in this way is because the current "legacy" handoff from BL2 and opteed is passing the AARCH mode via X0/R0, but not directly reading from SPSR. If you agree, I can remove this TE from the spec, and then modify the code to allow opteed to get the AARCH mode from SPSR instead of X0/R0 from BL2.

I see. Adding this explanation would help, i.e. the first sentence you wrote

danh-arm: > If you agree, I can remove this TE from the spec, and then modify the code to allow opteed to get the AARCH mode from SPSR instead of X0/R0 from BL2.

I think that's for the OP-TEE maintainers to agree, not the spec maintainers. Although it sounds tempting to read SPSR, I think AArch64 OPTEED may need to support launching AArch32 OP-TEE. Having said that, recent Arm CPUs don't support AArch32 at ELs > 0.

sjg20: Yes indeed, so just add the explanation and call it done?

raymo200915: > > If you agree, I can remove this TE from the spec, and then modify the code to allow opteed to get the AARCH mode from SPSR instead of X0/R0 from BL2.
> 
> I think that's for the OP-TEE maintainers to agree, not the spec maintainers. Although it sounds tempting to read SPSR, I think AArch64 OPTEED may need to support launching AArch32 OP-TEE. Having said that, recent Arm CPUs don't support AArch32 at ELs > 0.

Currently the opteed already holds the value of SPSR within the entry point info structure. So basically we can deprecate passing from the BL2 arguments. I already commit my changes for opteed and OP-TEE maintainers will review it.
For the spec, I think I can remove this TE from now.

source/transfer_list.rst
raymo200915: Oh, I get what you mean now. No, OP-TEE just needs the base address, it doesn't need the size of the area that this address pointed to.

raymo200915:
All comments have been addressed. Squashed and pushed the commit.

source/transfer_list.rst
raymo200915: Resolve this since the TE for AARCH is removed.

source/transfer_list.rst
jwerner-chromium: A once allocated TE may never be removed (at most it may be marked "deprecated"), so it would be better to try to make a final decision here right now.

raymo200915: Understand.
Then I will prefer to @jwerner-chromium 's suggestion, move OP-TEE TEs to start from 0x100 and keep all Arm/FF-A related TEs as "generic" ones start from 0.
Is this fine for you? @sjg20 @danh-arm

sjg20 (approved):
Yes that is OK, so long as the next project doesn't have to leave 256 positions!

raymo200915:
All comments have been addressed. Squashed and pushed the commit.

raymo200915:
> Just a few other minor comments

Comments are addressed.

source/transfer_list.rst
manish-pandey-arm: What about we keep truly generic ones like DTB/HOBLIST/TPM starting at 0x0 and let all Arm specific once starting 0x100. Arm specific once can include all the projects which are hosted in tf.org (TF-A/OP-TEE/FFA)

raymo200915: > What about we keep truly generic ones like DTB/HOBLIST/TPM starting at 0x0 and let all Arm specific once starting 0x100. Arm specific once can include all the projects which are hosted in tf.org (TF-A/OP-TEE/FFA)

This is fine for me. @sjg20 @danh-arm @jwerner-chromium , is this good for you? If all of you agree I will amend the PR.

sjg20: How many Arm-specific ones are we expecting to see there? I would like avoid a very sparse list. So long as expect several dozen over time, then that seems OK to me

sjg20 (approved):

source/transfer_list.rst
manish-pandey-arm: Thanks @raymo200915 !
To understand the usage of XFERLIST_OPTEE_PAGABLE_PART i was looking at your patch [1], this tag basically passes a base address and limit for the next stage for its usage. There is very similar usecase in bl1/bl2 interface where BL1 passes memory layout for BL2. Which basically a structure like below 
typedef struct meminfo {
        uintptr_t total_base;
        size_t total_size;
} meminfo_t;

I was just wondering if we could have a more generic name for this entry (XFERLIST_MEM_INFO or XFERLIST_MEM_LAYOUT) which can be used by receiving stage for its memory usage. This can even go in the generic range.

[1] https://review.trustedfirmware.org/c/TF-A/trusted-firmware-a/+/23777

raymo200915: @manish-pandey-arm
`XFERLIST_OPTEE_PAGABLE_PART` is passing the 'base address' of the OP-TEE pagable part only, not the memory range pointed by that address.
Other than that, if we use a generic name, then there might be multiple TEs with same tag even within one boot stage, how OP-TEE knows which one  is containing the pagable part to consume?

source/transfer_list.rst
```diff
@@ -657,5 +658,84 @@ such that the last ACPI table in this entry ends at offset
...
+
+   * - data_size
+     - 0x4
+     - 0x4
+     - The size of OP-TEE pagable part in bytes must be set to **8**.
+
+   * - pagable_part
+     - data_size
+     - hdr_size
+     - Pagable part address for OPTEED
```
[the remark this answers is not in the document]
raymo200915: The `Pagable part address for OPTEED` is a base address for OP-TEE to load from the memory for its 'pagable part'.
It is known as 'the base address of OP-TEE pagable part'.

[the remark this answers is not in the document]
sjg20: What is a pagable part? Does it mean that the memory pages in and out? I'm just unsure what this area is used for.

sjg20 (changes requested):

source/transfer_list.rst
```diff
@@ -657,5 +658,84 @@ such that the last ACPI table in this entry ends at offset
...
 
+.. _arm_entries:
+
+Arm specific entries
+^^^^^^^^^^^^^^^^^^^^
+
+The following entry types are defined for Arm projects,
+including TF-A, OP-TEE and FF-A:
+
+**OP-TEE pagable part entry layout (XFERLIST_OPTEE_PAGABLE_PART)**
```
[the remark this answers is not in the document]
raymo200915: I guess to make it more precise, we can use the term 'the base address of OP-TEE pagable part', however, it is a bit difficult to give more information about it since it should be described in the design document of OP-TEE and out of scope in this document. Maybe a link to the OP-TEE design page is a good idea.

source/transfer_list.rst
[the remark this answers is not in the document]
raymo200915: As mentioned above, it means 'the base address of OP-TEE pagable part' that for OP-TEE to load its pagable part.

source/transfer_list.rst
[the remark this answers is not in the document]
sjg20: A link would help, but I think this should explain here what a pagable part is. That is the bit I don't get. It sounds like it means something, but what? How does one choose the value to put here? Why does it have just an address, and not a size as well?

source/transfer_list.rst
sjg20: Yes I think it is good to have this in its own TE, so I don't agree with using generic mem info/layout for this.

But can you explain why the size is not required?

manish-pandey-arm: Ack, for having its own TE.

source/transfer_list.rst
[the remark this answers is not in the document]
raymo200915: Following the way we were introducing FDT and HOB, I added `[OPTEECore]` and `[ArmFFA]` links for reference.
I keep the term `OP-TEE pageable part address` to align with the name that is usually called and explain it as `the address of OP-TEE pageable part` in the descriptions and tables.
I think this is fair enough and not necessary to explain in this spec why only an address is needed or why the pointed memory is not required  - it will be too verbose and out-of-scope and they should be described in the related OP-TEE docs (Similar with what we did for FDT and HOB, we didn't even explain what they are in this spec).

raymo200915:
All comments have been addressed. A new commit is pushed.

sjg20 (approved):

source/transfer_list.rst
```diff
@@ -657,5 +658,86 @@ such that the last ACPI table in this entry ends at offset
      - hdr_size
      - One or more ACPI tables.
 
+.. _arm_entries:
+
+Arm specific entries
+^^^^^^^^^^^^^^^^^^^^
+
+The following entry types are defined for Arm projects,
+including TF-A, OP-TEE and FF-A:
```
danh-arm, line 667: Suggest "The following entry types are defined for TF-A related projects, including OP-TEE". I don't think FF-A should be mentioned here.

source/transfer_list.rst
```diff
@@ -657,5 +658,86 @@ such that the last ACPI table in this entry ends at offset
...
+     - 0x4
+     - 0x4
+     - The size (in bytes) of the address of OP-TEE pageable part which must be set to **8**.
+
+   * - pp_addr
+     - data_size
+     - hdr_size
+     - Holds the address of OP-TEE pageable part
+
+**SPMC manifest entry layout (XFERLIST_SPMC_MANIFEST)**
```
danh-arm, line 706: I think the way this is used in TF-A is more general than this. TF-A defines the concept of a FW_CONFIG. In the context of SPMC configs, TOS_FW_CONFIG is overloaded as the SPMC manifest. However, in other configs, this translates into a BL specific config using the DT format. So I think this should be something like XFERLIST_TFA_FW_CONFIG, and explain that the format of the fw_config is statically known by the receiver. For example, it may be an SPMC manifest or a receiver specific configuration image.

source/transfer_list.rst
```diff
@@ -657,5 +658,86 @@ such that the last ACPI table in this entry ends at offset
...
+     - 0x1
+     - 0x3
+     - |hdr_size_desc|
+
+   * - data_size
+     - 0x4
+     - 0x4
+     - The size of SPMC manifest in bytes.
+
+   * - spmc_manifest
```
danh-arm, line 737: fw_config

source/transfer_list.rst
```diff
@@ -657,5 +659,121 @@ such that the last ACPI table in this entry ends at offset
...
+
+   * - hdr_size
+     - 0x1
+     - 0x3
+     - |hdr_size_desc|
+
+   * - data_size
+     - 0x4
+     - 0x4
+     - The size of OP-TEE pagable part in bytes must be set to **8**.
```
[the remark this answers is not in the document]
raymo200915: Please check the latest commit, it should be:
```
   * - data_size
     - 0x4
     - 0x4
     - The size (in bytes) of the address of OP-TEE pageable part which must be set to **8**.
```
We are not passing the pageable part, but only the address of the pageable part.

source/transfer_list.rst
```diff
@@ -657,5 +658,86 @@ such that the last ACPI table in this entry ends at offset
      - hdr_size
      - One or more ACPI tables.
 
+.. _arm_entries:
+
+Arm specific entries
```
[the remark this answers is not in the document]
raymo200915: I am open for this naming, let's see if there comes more inputs from @manish-pandey-arm @sjg20 @jwerner-chromium

source/transfer_list.rst
raymo200915: SPMC manifest is a term from FF-A spec.

source/transfer_list.rst
raymo200915: But seems that `FW_CONFIG` is too general and actually it can mean anything. For example, even FDT and HOB are some kind of `FW_CONFIG`. This is not good for a TE's definition.

source/transfer_list.rst
raymo200915: Same comment as above.

source/transfer_list.rst
manish-pandey-arm: > But seems that `FW_CONFIG` is too general and actually it can mean anything. For example, even FDT and HOB are some kind of `FW_CONFIG`. This is not good for a TE's definition. I think better to define different TEs in the future for different receivers and purposes instead of just a general `FW_CONFIG`.

Agree, I also have same thoughts !

danh-arm: Yes, FW_CONFIG is a general concept but that is how the code is written today so my suggestion is to just to align the spec with the code. It's true the definition of what's in the FW_CONFIG is then outside the spec but it at least provides a solution for passing all types of FW_CONFIG, not just this specific SPMC manifest case.
We can discuss changing TF-A to define FW_CONFIGs with specific content for each case (instead of DT) but I don't think this PR is the best place for that conversation. That would need a separate PR anyway.

danh-arm (changes requested):

source/transfer_list.rst
raymo200915: Once a TE entry is defined it cannot be changed anymore due to backward compatibility, I think now it is the time to define the new TEs with their real purposes instead of something general - Even right now the code is still using the old concepts.
If we still define `FW_CONFIG` here, we will have to deprecate it in a near future.

raymo200915: I think now the question is whether we should keep the spec to reflect to the current code or the spec should move ahead from the code? Hmmm... I prefer to the latter.

danh-arm: It's hard to change an existing TE but it's easy to create new ones and use them instead. However, I agree it would be better to align with where we want to go rather than change later. I personally don't see a strong need to change from the existing FW_CONFIG - it makes it easy to extend without requiring spec changes. I will talk to @manish-pandey-arm any other to at least get an agreed Arm position.

danh-arm: > Dan are you suggesting that we might add new members to this struct later, increasing its size?

No, we're just discussing whether to overload the meaning of the data in this TE or if there's a risk of us changing the meaning in future.

I discussed this more with @manish-pandey-arm. Manish wants to explore passing more specific TEs for some BLs/configs to avoid the overhead of DT parsing in some images. However, we think are some configs/platforms where it will always be necessary to pass some kind of BL-specific config in device tree format separate to XFERLIST_FDT. A generic fw_config TE could be a solution to that, but that does not need to be addressed in this PR. While such a TE could potentially also be used for this SPMC manifest, I'm now happy for a separate TE to be defined for this. It might help with parsing in the producer, especially if additional BL-specific config data needs to be passed later.

I have a few more questions/comments:

- According to the FF-A spec, the format of the SPMC manifest is implementation defined. I believe you're intending to use this in TF-A where the DT format is used, right? In which case, the DT format should be specified here and the name should specify DT to enable other SPMC formats, e.g. XFERLIST_DT_SPMC_MANIFEST.

- The FF-A spec defines it's own handoff protocol (called FF-A boot protocol), which competes with this spec. Is it a conscious decision to use this spec instead of the FF-A boot protocol? I'd say that's good idea anyway as this spec is likely to be more dynamic that the FF-A spec.

- If this was an FF-A TE, then I would say it should be in a separate (e.g. Arm architecture) range of TEs. However, if my above comment is correct, this is really still a TF-A TE since TF-A chooses the SPMC manifest format.

raymo200915: @danh-arm Seems to be a reasonable solution.
I will rename the TE tag to `XFERLIST_DT_SPMC_MANIFEST` and the group name to `TF-A related`.
Link to 'https://trustedfirmware-a.readthedocs.io/en/latest/components/secure-partition-manager.html#spmc-manifest' instead of FF-A spec.

source/transfer_list.rst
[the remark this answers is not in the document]
manish-pandey-arm: Agree to Dan's suggestion with minor nit "Trusted Firmware related entries". 

https://www.trustedfirmware.org/ has various projects (including TF-A, OP-TEE, TF-M, MCUboot etc). Even TF-M might use it in future. My preference is to avoid "-A", if it is usable for M class projects as well.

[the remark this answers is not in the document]
raymo200915: Fixed.

source/transfer_list.rst
raymo200915: Fixed.

source/transfer_list.rst
raymo200915: Done.

source/transfer_list.rst
raymo200915: Fixed.

raymo200915:
All comments have been addressed. Please review.

source/transfer_list.rst
```diff
@@ -657,5 +658,86 @@ such that the last ACPI table in this entry ends at offset
      - hdr_size
      - One or more ACPI tables.
 
+.. _tf_entries:
+
+Trusted Firmware related entries
```
sjg20, line 663: Trusted Firmware-related entries

or

Entries related to Trusted Firmware

sjg20 (approved):

source/transfer_list.rst
```diff
@@ -657,5 +658,86 @@ such that the last ACPI table in this entry ends at offset
...
+
+   * - pp_addr
+     - data_size
+     - hdr_size
+     - Holds the address of OP-TEE pageable part
+
+**DT formatted SPMC manifest entry layout (XFERLIST_DT_SPMC_MANIFEST)**
+
+This entry type holds the SPMC (Secure Partition Manager Core) manifest image
+which is in DT formatt [DT]_ and described in [TFASPM]_.
```
danh-arm, line 709: minor: s/formatt/format/

source/references.rst
```diff
@@ -8,3 +8,7 @@
 .. [ACPI] `Advanced Configuration and Power Interface (ACPI) Specification.`
 
 .. [ArmARM] `Arm® Architecture Reference Manual for the A-profile architecture`, Arm LTD.
+
+.. [OPTEECore] `OP-TEE Core <https://optee.readthedocs.io/en/latest/architecture/core.html>`
+
+.. [TFASPM] `TF-A SPM <https://trustedfirmware-a.readthedocs.io/en/latest/components/secure-partition-manager.html>`
```
danh-arm, line 14: A better reference seems to be this DT binding doc:
https://trustedfirmware-a.readthedocs.io/en/latest/components/ffa-manifest-binding.html

Also, these hyperlinks don't seem to be clickable in the rendered version.

danh-arm (dismissed):
My comments are only minor so I'll approve this. But it would be nice to fix them.

source/transfer_list.rst
raymo200915: Fixed.

source/transfer_list.rst
raymo200915: Fixed.

raymo200915:
Fixed a few minor issues commented by @danh-arm and @sjg20 .

source/references.rst
danh-arm: I meant the hyperlinks within the references (see the Bibliography section). But I'm not sure if RST supports this and we can look at this later.

danh-arm (approved):

PR #28, opened 2023-11-16
the table is not versioned, the document is

thomas-fossati:


sjg20 (approved):
I'm not sure this makes much different, but OK

jmarinho (commented):
We should be cautious here.

The TL header is unlikely to change.
Unless we need to append new fields to it, or (even less likely) change the semantics of some exiting field, we should not change its version.

As for the document, I assume we bump its version at every release.
The TL version has a single byte.

I'm of the opinion that there ought to be a version for the TL header and another for the document.

thomas-fossati:
It's not clear what _"this version of the table"_ refers to since the table isn't explicitly versioned, and it doesn't seem to me that it's published independently from the document.

If the document is expected to undergo frequent changes, it might be better to separate the bit containing the core specification from the extensions registry.

danh-arm (changes requested):
This PR should be rejected. The version absolutely applies to the TL, not the document. The current version of the doc is 0.9 and will be incremented independently to the version of the TL (currently 0x1). For example, when we add new TE types and want to make a release, we will increment the document version but we will not increment the TL header.

thomas-fossati:
> This PR should be rejected. The version absolutely applies to the TL, not the document.

The problem is the document currently defines the _"version of the table"_ as the anchor point for the described header syntax and semantics.  However, there is no such thing as a "table version."  Besides, it's not entirely clear what the table in question is.

The text in the PR just moves the anchoring point to the document, irrespective of its version.

(See also https://github.com/FirmwareHandoff/firmware_handoff/pull/28#issuecomment-1816838332.)

danh-arm:
> there is no such thing as a "table version." 

OK, strictly speaking there's no "table version" but the intent is that this refers to the layout corresponding to the current version of the TL header.

> The text in the PR just moves the anchoring point to the document, irrespective of its version.

But it's important that we refer to the current version (of the TL header), because the values may be different in future versions of the TL header.

How about this language instead?
"This field is set to X for the layout described in this version of the TL header."

thomas-fossati:
> But it's important that we refer to the current version (of the TL header), because the values may be different in future versions of the TL header.
> 
> How about this language instead? "This field is set to X for the layout described in this version of the TL header."

Since there is no other TL header layout described in the document, there would be no ambiguity just saying "This field is set to X for the layout described in this document." (which is what the PR does).

BTW, apologies for the pedantry!

danh-arm:
> Since there is no other TL header layout described in the document...

There may be many versions of the document out there, each with a different TL header layout. It's critical that implementers of particular TL header version are referring to a document with the same TL header version.

> BTW, apologies for the pedantry!

No apology needed but I'd argue that I (and the original author) are using more pedantic language, which is why I'd rather not water this down with the language you propose.

I don't want to spend any more time on this one so could you please update as suggested (or similar) or withdraw this PR?

PR #29, opened 2023-11-16
remove spurious NL

thomas-fossati:


sjg20 (approved):

jmarinho (approved):

danh-arm (approved):
OK, but I'm not a fan of trivial PRs like this. Better to do as part of a more significant change.

PR #31, opened 2023-12-21
Add executable entrypoint information TE entry

harrisonmutai-arm:
Add a new TE entry type for executable entry point information to TF-A's entry range. This entry type contains data about the execution state of images (i.e. PC, SPSR values), which can be used by a consumer to execute those images.

### DIFF
--- a/source/references.rst
+++ b/source/references.rst
@@ -26,3 +26,5 @@ References
 .. [TFAFFAMB] TF-A Secure Partition Manager: `FF-A manifest binding to device tree <https://trustedfirmware-a.readthedocs.io/en/latest/components/ffa-manifest-binding.html>`__
 
 .. [TCG_EFI] `Trusted Computing Group EFI Protocol Specification <https://trustedcomputinggroup.org/resource/tcg-efi-protocol-specification>`__
+
+.. [TF_BL31] `Data structures used in the BL31 cold boot interface <https://trustedfirmware-a.readthedocs.io/en/latest/design/firmware-design.html#data-structures-used-in-the-bl31-cold-boot-interface>`
--- a/source/references.rst
+++ b/source/references.rst
@@ -27,4 +27,4 @@ References
 
 .. [TCG_EFI] `Trusted Computing Group EFI Protocol Specification <https://trustedcomputinggroup.org/resource/tcg-efi-protocol-specification>`__
 
-.. [TF_BL31] `Data structures used in the BL31 cold boot interface <https://trustedfirmware-a.readthedocs.io/en/latest/design/firmware-design.html#data-structures-used-in-the-bl31-cold-boot-interface>`
+.. [TF_BL31] `Data structures used in the BL31 cold boot interface <https://trustedfirmware-a.readthedocs.io/en/latest/design/firmware-design.html#data-structures-used-in-the-bl31-cold-boot-interface>`__
--- a/source/transfer_list.rst
+++ b/source/transfer_list.rst
@@ -902,6 +902,54 @@ This manifest contains the SPMC attribute node consumed by the SPMD
      - hdr_size
      - Holds a SPMC manifest image in DT format.
 
+**AArch64 executable entry point information (XFERLIST_EXEC_EP_INFO64)**
+
+This entry type holds the AArch64 variant of `entry_point_info`.
+`entry_point_info` is a TF-A-specific data structure [TF_BL31]_ used to
+represent the execution state of an image; that is, the state of general purpose
+registers, PC, and SPSR.
+
+This information is used by clients to setup the execution environment of
+subsequent images. A concrete example is the execution of a bootloader such as
+U-Boot in non-secure mode. In TF-A, the runtime firmware BL31 uses an
+`entry_point_info` structure corresponding to the bootloader, to setup the
+general and special purpose registers. Following conventions
+outlined in :ref:`aarch64_receiver`, the general purpose registers consumed
+by the bootloader contain the base addresses of the device tree, and transfer
+list; along with the transfer list signature.
+
+In practice, control might be transferred from BL31 to any combination of
+software running in Secure, Non-Secure, or Realm modes.
+
+.. _tab_entry_point_info:
+.. list-table:: Entry point info type layout
+   :widths: 2 5 2 6
+
+   * - Field
+     - Size (bytes)
+     - Offset (bytes)
+     - Description
+
+   * - tag_id
+     - 0x3
+     - 0x0
+     - The tag_id field must be set to **0x102**.
+
+   * - hdr_size
+     - 0x1
+     - 0x3
+     - |hdr_size_desc|
+
+   * - data_size
+     - 0x4
+     - 0x4
+     - Size of the `entry_point_info` structure in bytes.
+
+   * - ep_info
+     - `sizeof(entry_point_info)`
+     - hdr_size
+     - Holds a single `entry_point_info` structure.
+
 **Read-Write Memory Layout Entry Layout (XFERLIST_RW_MEM_LAYOUT64)**
 
 This entry type holds a structure that describes the layout of a read-write


source/transfer_list.rst
```diff
@@ -902,5 +902,42 @@ This manifest contains the SPMC attribute node consumed by the SPMD
      - hdr_size
      - Holds a SPMC manifest image in DT format.
 
+
+**Executable entry point information entry layout (XFERLIST_EXEC_EP_INFO)**
+
+This entry type holds an array of `entry_point_info_t` structures.
+`entry_point_info_t` is a data structure containing information used by a
```
danh-arm, line 909: I think we need a little more info here. This range is used for all TF projects and is not exclusive to TF-A, so you should say here that this is specifically a TF-A project structure. Also, it would be good to say this contains info like the PC, SPSR, GP regs (like you do in the commit message).

Ideally (but not essential for this PR), it would be nice to refer to a TF-A document that explains how BL31 uses this entry_point_info_t array to execute the bootflow.

danh-arm (changes requested):
Mostly good. Just a little more info needed

source/transfer_list.rst
harrisonmutai-arm: I've added a reference to [1](https://trustedfirmware-a.readthedocs.io/en/latest/design/firmware-design.html#data-structures-used-in-the-bl31-cold-boot-interface) which explains at a high level how BL31 uses `entry_point_info_t` in the cold boot flow, and added details about its contents i.e. register representation.

source/transfer_list.rst
```diff
@@ -906,9 +906,10 @@ This manifest contains the SPMC attribute node consumed by the SPMD
 **Executable entry point information entry layout (XFERLIST_EXEC_EP_INFO)**
 
 This entry type holds an array of `entry_point_info_t` structures.
-`entry_point_info_t` is a data structure containing information used by a
-bootloader to initialize the execution state of an image. This entry provides
-the consumer with all information to execute subsequent images.
+`entry_point_info_t` is a TF-A specific data structure [TF_BL31]_ used to
```
sjg20, line 909: TF-A-specific

source/transfer_list.rst
```diff
@@ -902,5 +902,42 @@ This manifest contains the SPMC attribute node consumed by the SPMD
...
+
+   * - data_size
+     - 0x4
+     - 0x4
+     - The size of SPMC manifest in bytes.
+
+   * - ep_info
+     - data_size
+     - hdr_size
+     - Holds an array of `entry_point_info_t` structures.
```
sjg20, line 941: This struct should be defined here. The spec should stand along, so we need to know what is in it.

Also, why do we want an array of these? I suspect that is fine, but I just don't know why...

harrisonmutai-arm: There are boot-loaders responsible for the execution of more than one image. For example, BL31 in some configurations will execute BL32, RMM and a non-secure image. Packing the information in an array, allows several entry points to be shared with the next stage in a single TE. We would otherwise require several TE's to convey the same information.

sjg20: Probably using several TEs is better

source/transfer_list.rst
```diff
@@ -906,9 +906,26 @@ This manifest contains the SPMC attribute node consumed by the SPMD
...
-bootloader to initialize the execution state of an image. This entry provides
-the consumer with all information to execute subsequent images.
+`entry_point_info_t` is a TF-A-specific data structure [TF_BL31]_ used to
+represent the execution state of an image, that is, the state of general purpose
+registers, PC and SPSR.
+
+.. code:: c
+
+    typedef struct entry_point_info {
+        param_header_t h;
```
sjg20, line 916: Can you please define all these types - param_header_t  aapcs64_params_t etc.

also please drop the 'typedef'

source/transfer_list.rst
```diff
@@ -906,9 +906,26 @@ This manifest contains the SPMC attribute node consumed by the SPMD
...
+        uintptr_t pc;
```
sjg20, line 917: uintptr_t can change size depending on the machine word size. Perhaps you need to use two different tags, depending on the architecture? Or use uint32_t here?

source/transfer_list.rst
```diff
@@ -906,9 +906,26 @@ This manifest contains the SPMC attribute node consumed by the SPMD
...
+        uint32_t spsr;
+    #ifdef __aarch64__
+        aapcs64_params_t args;
```
sjg20, line 920: Does this change the size of the struct?

sjg20 (changes requested):

source/transfer_list.rst
harrisonmutai-arm: As this architecturally defined i.e. 8 bytes wide on Aarch64, and 4 on Aarch32 makes most sense to define two separate tags.

source/transfer_list.rst
harrisonmutai-arm: Yes it does, added definitions for both structures below for clarity.

source/transfer_list.rst
harrisonmutai-arm: @raymo200915 raised this in the corresponding TF-A patch. Does the specification allow for several entries of the same TE type in a TL?

raymo200915: The spec doesn't tell explicitly whether multiple TEs can share same tag ID or I am wrong, but tag ID is the only field that can identify a TE at the moment.
Shall we consider to add another field in the TE header, so that we can identify a TE even when its tag ID is not unique in a TL?

manish-pandey-arm: > The spec doesn't tell explicitly whether multiple TEs can share same tag ID or I am wrong, but tag ID is the only field that can identify a TE at the moment. 
Even though spec does not explicitly mentions it, IMHO we should avoid using same tag ID for multiple TEs. This will keep things simple (e.g the searching of a tag ID returns the first occurrence) 

> Shall we consider to add another field in the TE header, so that we can identify a TE even when its tag ID is not unique in a TL?
IMO we should avoid adding a new field just to cater for a specific case (until its unavoidable).  As we know that there are going to be finite number of TEs (less than 10) and we have enough range to accommodate it in tag ID.
In this particular case the maximum array size would be 3 (BL32/BL33, RMM). So even if we decide to have several TEs we can easily provide 3 tag IDs in Arm range.

source/transfer_list.rst
```diff
@@ -902,5 +902,97 @@ This manifest contains the SPMC attribute node consumed by the SPMD
...
+
+   * - hdr_size
+     - 0x1
+     - 0x3
+     - |hdr_size_desc|
+
+   * - data_size
+     - 0x4
+     - 0x4
+     - The size of SPMC manifest in bytes.
```
manish-pandey-arm, line 990: SPMC manifest??

source/transfer_list.rst
jwerner-chromium: > The spec doesn't tell explicitly whether multiple TEs can share same tag ID

It does (now) actually, see the "TL Contents" section. I don't think it forbids putting an array in a single tag, though. In my opinion, this probably depends on whether these are generated all in one go or a different points in time. If you know for sure that the passing stage will always add all of these entry points at once, I think an array would be okay. If you think there's a chance that one stage might add some entry points and then later either the same stage or a later one might want to add some more, then using a separate TE for each would be a good idea.

I'm not really sure what you mean by "identify a TE"? Within the context of a single firmware stage, a TE can be identified by its address, of course (e.g. to pass around between APIs). Across firmware stages I don't see why we would want to identify TEs, since every TE should be independent and not reference other TEs. When parsing a TL to find something (e.g. your entry points here), I would suggest an API that searches for all TEs of a certain tag ID and then returns each of them in turn (e.g. with a callback or a find_next() function) so the caller can figure out which one(s) it is interested in.

manish-pandey-arm: >  this probably depends on whether these are generated all in one go or a different points in time

In context of current PR, these entries are all generated at once in BL2 stage.

raymo200915: I would prefer to have different TE with different ID for each ep_info instead of putting all ep_info into an array, just in order to simply the scenario in BL31 - if we still have sufficient spare IDs.

> I'm not really sure what you mean by "identify a TE"? 
I mean the operations to locate a specific TE. For example, search a TE by a tag ID or anything else and returns an TE address. Anyway, if we are not going to have multiple TEs with same tag ID, we don't have the worries.

ManishVB-Arm: Chapter 'TL Contents'  has below statement -

> Tags are expected to have a simple layout (representable by a C structure) and each tag should only 
> represent data for a single logical concept. Data for multiple distinct concepts should be split across
> different tags

IIUC, it signifies that a single or multiple sets of data can be denoted by the same tag-ID, as long as they encapsulate the same logical entity. In this specific scenario, the entry point is a singular logical entity, thus portraying it in either a singular (single entry-point info) or multiple fashion (multiple entry-point infos) should be considered acceptable.
If above understanding is correct then presenting same logical information through multiple tags could potentially create confusion.

harrisonmutai-arm: If re-usability is a stated design goal of the specification, our options are either to maintain the TE as described in this patch, or to more narrowly define it so that each instance of the structure gets its own entry. From @ManishVB-Arm message it seems either is viable, although, some rework might be required in the implementation used by TF-A to support having multiple instances of the same TE. 
>The same tag ID may occur multiple times in the TL to represent multiple instances of the same kind of object.

On the other suggestion, I think it would be misguided to create tags for each type of `entry_point_info_t`. Having multiple representations of the same data risks creating confusion around usage. This could result in a scenario where every platform/chipset has to re-invent the wheel for their uses-case because of a lack of a generic representation.

source/transfer_list.rst
```diff
@@ -907,23 +907,61 @@ This manifest contains the SPMC attribute node consumed by the SPMD
...
+registers, PC, and SPSR.
 
 .. code:: c
 
     typedef struct entry_point_info {
         param_header_t h;
-        uintptr_t pc;
         uint32_t spsr;
+
     #ifdef __aarch64__
```
sjg20, line 919: Perhaps I am becoming confused by github's awful code-review system, but it looks like you still have the #ifdef in here. Instead, these two structures should be defined separately

source/transfer_list.rst
```diff
@@ -907,23 +907,61 @@ This manifest contains the SPMC attribute node consumed by the SPMD
...
 This entry type holds an array of `entry_point_info_t` structures.
 `entry_point_info_t` is a TF-A-specific data structure [TF_BL31]_ used to
-represent the execution state of an image, that is, the state of general purpose
-registers, PC and SPSR.
+represent the execution state of an image; that is, the state of general purpose
+registers, PC, and SPSR.
 
 .. code:: c
 
     typedef struct entry_point_info {
```
sjg20, line 915: please void typedef. We actually are not currently using C in the spec, so things are written out in table form.

However, I would be OK with us using C instead, if others agree. But perhaps cross that bridge later, after this PR is applied?

sjg20 (changes requested):

source/transfer_list.rst
```diff
@@ -907,23 +907,61 @@ This manifest contains the SPMC attribute node consumed by the SPMD
...
+
+    struct param_header {
+        uint8_t type;   /* type of the structure */
+        uint8_t version;  /* version of this structure */
+        uint16_t size;    /* size of this structure in bytes */
+        uint32_t attr;    /* attributes: unused bits SBZ */
+    } param_header_t;
+
+General purpose register state is stored in the `_params` field. The size and width
+of these registers is architecturally defined.
```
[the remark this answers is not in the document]
harrisonmutai-arm: I've converted these C code blocks to tables and added some details about how `entry_point_info_t` should be used i.e. registers values are populated following handoff convention with tl information and pointer to device tree if there is one.

source/transfer_list.rst
harrisonmutai-arm: @sjg20 4-bytes of `param_header_t` are reserved to provide the attributes of the represented structure. In fact, we already use bits 0 and 5 of that field to encode the security state. At the moment, I'm using those bits as a proxy for the image id. That is enough for the reference platform I'm working on, although, there's no reason we can't add an additional image id field in the attributes.

source/transfer_list.rst
```diff
@@ -902,5 +902,172 @@ This manifest contains the SPMC attribute node consumed by the SPMD
...
+
+   * - hdr_size
+     - 0x1
+     - 0x3
+     - |hdr_size_desc|
+
+   * - data_size
+     - 0x4
+     - 0x4
+     - Size of `entry_point_info_t` structures in bytes.
```
raymo200915, line 1065: Size of multiple `entry_point_info_t` if any?

source/transfer_list.rst
```diff
@@ -902,5 +902,172 @@ This manifest contains the SPMC attribute node consumed by the SPMD
...
+     - 0x3
+     - Size of `param_header_t`.
+
+   * - attr
+     - 0x4
+     - 0x5
+     - Attributes of the represented structure.
+
+.. param_header_attr_state:
+.. list-table:: Security state encoding in `param_header_t` attributes for `entry_point_info_t`.
```
raymo200915, line 1015: Does this mean we would like to use the security state instead of an index or ID when picking up an `entry_point_info_t` from multiple instances if there are any?

source/transfer_list.rst
```diff
@@ -902,5 +902,172 @@ This manifest contains the SPMC attribute node consumed by the SPMD
...
+
+   * - data_size
+     - 0x4
+     - 0x4
+     - Size of `entry_point_info_t` structures in bytes.
+
+   * - ep_info
+     - data_size
+     - hdr_size
+     - Holds one or more `entry_point_info_t` structure.
```
raymo200915, line 1070: In an array of structure?

source/transfer_list.rst
```diff
@@ -902,5 +902,42 @@ This manifest contains the SPMC attribute node consumed by the SPMD
...
+     - 0x3
+     - 0x0
+     - The tag_id field must be set to **0x102**.
+
+   * - hdr_size
+     - 0x1
+     - 0x3
+     - |hdr_size_desc|
+
+   * - data_size
```
sjg20, line 1062: Is this needed? It looks like the value is always 8 so it can be omitted.

source/transfer_list.rst
```diff
@@ -902,5 +902,42 @@ This manifest contains the SPMC attribute node consumed by the SPMD
...
+     - 0x4
+     - 0x4
+     - The size of SPMC manifest in bytes.
+
+   * - ep_info
+     - data_size
```
sjg20, line 1068: the DT takes up the whole entry, so we should be able to drop data_size

source/transfer_list.rst
```diff
@@ -902,5 +902,172 @@ This manifest contains the SPMC attribute node consumed by the SPMD
...
+
+   * - attr
+     - 0x4
+     - 0x5
+     - Attributes of the represented structure.
+
+.. param_header_attr_state:
+.. list-table:: Security state encoding in `param_header_t` attributes for `entry_point_info_t`.
+
+   * - Bit 5
```
sjg20, line 1017: Can we use consecutive bits (0 and 1) ?

source/transfer_list.rst
```diff
@@ -902,5 +902,172 @@ This manifest contains the SPMC attribute node consumed by the SPMD
...
+   * - 1
+     - 1
+     - Realm
+
+This information is used by the receiver to setup the execution environment of
+subsequent images. A concrete example is the execution of a bootloader in
+Non-Secure mode i.e. U-Boot. In TF-A, the runtime firmware BL31 uses an
+`entry_point_info_t` structure corresponding to the bootloader, to setup the
+general and special purpose registers. The general purpose registers consumed by
+bootloader contain the base address of the hardware description device tree,
```
sjg20, line 1038: Thanks for adding the extra info.

instead of 'hardware-description device tree' can we just save device tree? The devicetree is provided by the XFERLIST_FDT entry

source/transfer_list.rst
```diff
@@ -902,5 +902,172 @@ This manifest contains the SPMC attribute node consumed by the SPMD
...
 
+
+**Executable entry point information entry layout (XFERLIST_EXEC_EP_INFO)**
+
+This entry type holds one or more `entry_point_info_t` structures.
+
+`entry_point_info_t` is a TF-A-specific data structure [TF_BL31]_ used to
+represent the execution state of an image; that is, the state of general purpose
+registers, PC, and SPSR. Due to architectural differences in AArch64 and
+AArch32, there exist two representations of `entry_point_info_t`.
```
sjg20, line 913: In that case, how about having two different tags? It will be difficult for a reader to validate the entries if the description for  a tag two options.

source/transfer_list.rst
```diff
@@ -902,5 +902,172 @@ This manifest contains the SPMC attribute node consumed by the SPMD
...
+
+.. param_header_t:
+.. list-table:: Layout of header data structure `param_header_t`.
+
+   * - Field
+     - Size (bytes)
+     - Offset (bytes)
+     - Description
+
+   * - type
```
sjg20, line 993: Is this the security state? Can we just call this 'security state'?

source/transfer_list.rst
```diff
@@ -902,5 +902,172 @@ This manifest contains the SPMC attribute node consumed by the SPMD
...
+     - 0x1
+     - 0x0
+     - Type of the represented structure i.e. executable info.
+
+   * - version
```
sjg20, line 998: We should not need a version within a TE. Can we drop this field?

source/transfer_list.rst
```diff
@@ -902,5 +902,172 @@ This manifest contains the SPMC attribute node consumed by the SPMD
...
+     - 0x1
+     - 0x1
+     - Version of the `param_header_t` structure (version 2 in this
+       specification).
+
+   * - size
+     - 0x2
+     - 0x3
+     - Size of `param_header_t`.
```
sjg20, line 1007: The size appears to be fixed, so we should be able to drop this field. The TE already has a size

source/transfer_list.rst
sjg20: This starts at offset 5 and is size ...so does it mean the header is 9 bytes? Can we use an aligned size?

Also, where is attr defined?

source/transfer_list.rst
```diff
@@ -902,5 +902,172 @@ This manifest contains the SPMC attribute node consumed by the SPMD
      - hdr_size
      - Holds a SPMC manifest image in DT format.
 
+
+**Executable entry point information entry layout (XFERLIST_EXEC_EP_INFO)**
+
+This entry type holds one or more `entry_point_info_t` structures.
```
sjg20, line 908: I believe it would be better to put each structure in its own TE, rather than packing multiple records into a TE, which we try to avoid. From what I can tell, there is one record per security state, with the 'type' field indicating the state?

sjg20 (changes requested):

source/transfer_list.rst
danh-arm: I tend to agree it would be more flexible for the TE to represent a single entry_point_info but I don't feel strongly about it. However, if we do that, I do feel strongly that the same tag_id should be used for all entries and that there should _not_ be a separate id field, since this would be duplicate information. Parsers can distinguish the entries by parsing all the TEs of this type.
By the way, the type is indicated by the attr field of param_header, not the type field (which tells you that this is an entry_point_info structure). param_header is used for other structures too.

source/transfer_list.rst
danh-arm: This is a long-standing external interface structure in TF-A. Maintainers of this spec have no business requesting changes to it.

source/transfer_list.rst
danh-arm: This is a long-standing external interface structure in TF-A. Maintainers of this spec have no business requesting changes to it.

source/transfer_list.rst
danh-arm: This is a long-standing external interface structure in TF-A. Maintainers of this spec have no business requesting changes to it. Having said that, there's a mistake here; size is at offset 0x2 and attr is at offset 0x4.
Yes, we use the security state to distinguish between multiple entry_point_info instances. If we need to further distinguish these in future (e.g. for multiple instances for the same security state), we can add/modify a field at that time.

source/transfer_list.rst
danh-arm: This is a long-standing external interface structure in TF-A. Maintainers of this spec have no business requesting changes to it. The bits in between are used for other stuff not relevant to this TE.

source/transfer_list.rst
danh-arm: This is a variable length inline array, not a pointer.

source/transfer_list.rst
danh-arm: Not sure what you mean. data_size is a standard TE header field. The ep_info array is of variable size and even if it wasn't, data_size enables us to add new fields later.

source/transfer_list.rst
```diff
@@ -902,5 +902,172 @@ This manifest contains the SPMC attribute node consumed by the SPMD
...
+     - 0
+     - Secure
+
+   * - 0
+     - 1
+     - Non-Secure
+
+   * - 1
+     - 1
+     - Realm
```
danh-arm, line 1031: I think it's a mistake for us to explicitly specify entry_point_info and param_header header. We don't do it for any other entries (e.g. DT, HOB, SPMC manifest, TPM event log, ....) so why start now? This stuff is already specified in the external reference: https://trustedfirmware-a.readthedocs.io/en/latest/design/firmware-design.html#data-structures-used-in-the-bl31-cold-boot-interface

Admittedly, that doc could do with more detail but I think the place for that detail is there, not here. Note that entry-point_info is itself a versioned and self-describing structure so can change in future without affecting compatibility in this spec.

danh-arm (changes requested):
Sorry, I've not been following this PR closely. I think it's gone in the wrong direction a bit. Maybe a quick design review is needed?

source/transfer_list.rst
sjg20: Oh right, but then why is the size variable? Shouldn't it always be 4? If this is extended in future, then we would want a standard size for spmc_man

source/transfer_list.rst
sjg20: Agreed that there shouldn't be a separate id field, since there seems to be a field indicating the type

source/transfer_list.rst
sjg20: Is this intended for use by multiple projects, or just one? Given that things are moving to firmware handoff, would that not be an opportunity to review the structure, since the generator and any clients would have to change anyway?

But if it is a pre-existing binary structure that we have no control over, then yes I suppose we just have to live with it.

Even so, we do need to think about the 32/64-bit side of things, as I commented elsewhere.

source/transfer_list.rst
danh-arm: I'm a bit confused by the mention of spmc_man when this is about ep_info. The size of a single ep_info can be discovered but this is an array of variable size and data_size enables us to determine how many entries are in the array. However, by using data_size in this way, it prevents us adding new fields to this TE in future. So I'm now thinking it might be better to add an "ep_info array size" field to enable future extension.

source/transfer_list.rst
harrisonmutai-arm: Let's remove the these descriptions if everyone is in agreement. Longer term, it will create a maintenance burden on this project if the information is duplicated here. I've raised a separate patch to capture the structure descriptions in TF-A's documentation: https://review.trustedfirmware.org/c/TF-A/trusted-firmware-a/+/26233.

source/transfer_list.rst
harrisonmutai-arm: > I'm a bit confused by the mention of spmc_man when this is about ep_info.

I think the confusion stems from an earlier diff that the original comment was applied to, this referred to spmc_man rather than ep_info.

> However, by using data_size in this way, it prevents us adding new fields to this TE in future. So I'm now thinking it might be better to add an "ep_info array size" field to enable future extension.

To address the issue of versioning, could a consumer of the TE not use type punning to interpret the the first header in the blob? By doing this, they have access to all the information they need to handle different versions of the data structure, including its size. Admittedly, this might introduce some additional complexity on the consumer side.

danh-arm: Yes, clients can inspect the ep_info header to determine its size (for a given version), but that won't tell you how many elements are in the array. This can be inferred from data_size if there are no other fields but if we add new fields later, an old consumer might try to interpret those new fields as another ep_info. I suppose it might be OK if the sizeof (new fields)  < sizeof(ep_info) but it's a bit flaky to rely on that.

source/transfer_list.rst
manish-pandey-arm: Agree, lets remove the description from here.

manish-pandey-arm (dismissed):

source/transfer_list.rst
harrisonmutai-arm: > but if we add new fields later, an old consumer might try to interpret those new fields as another ep_info.

That's a good point. If I’m not mistaken, you’re saying that we may at some point want to populate the data field with some additional information similar to the flags field `XFERLIST_EVLOG`? In which case, the data_size will not be a proxy variable for the ep_info count.

sjg20: Instead of count, how about having a separate TE for each piece of info? The contents should have enough info to know which state it relates to, right?

sjg20 (changes requested):

source/transfer_list.rst
jwerner-chromium: If everyone thinks writing parsers/APIs that can deal with multiple TEs which have the same tag is too complex, we should maybe change the spec to disallow them after all. It wouldn't be good if some parts of the ecosystem can support and use those constructs and others always try to find an alternative solution because their TL code can't handle it. I originally wrote the section that says it's allowed, but I admit I can't really come up with many practical examples where an array of structs within a single TE wouldn't do just as well. There may be the occasional edge case where a TE needs to be moved and expanded to add more entries to the array, but it should be pretty rare and maybe the ease of use of supporting a "get entry by tag ID" API is worth that.

source/transfer_list.rst
```diff
@@ -902,5 +902,59 @@ This manifest contains the SPMC attribute node consumed by the SPMD
...
+     - 0x1
+     - 0x3
+     - |hdr_size_desc|
+
+   * - data_size
+     - 0x4
+     - 0x4
+     - Size of `entry_point_info` structures in bytes.
+
+   * - count
```
harrisonmutai-arm, line 949: @danh-arm this seemed like the most sensible place to put the count, but I'm slightly unsure whether it is the right thing to do since it complicates how we access the array on the client side. Does this make sense?

source/transfer_list.rst
harrisonmutai-arm: After discussing with Manish and experimenting with our library implementation, we’ve decided to adopt the single EP per TE approach proposed by @sjg20. It appears that managing multiple TEs with the same tag ID in the transfer list is simpler than initially anticipated. Overall, packing the structures into an array introduces complexity for both the sender and receiver, especially if we need to include additional metadata solely to convey information about the array.

source/transfer_list.rst
```diff
@@ -902,5 +902,54 @@ This manifest contains the SPMC attribute node consumed by the SPMD
      - hdr_size
      - Holds a SPMC manifest image in DT format.
 
+
+**Executable entry point information entry layout (XFERLIST_EXEC_EP_INFO)**
+
+This entry type holds an `entry_point_info` structure.
```
danh-arm, line 908: As Simon said in an earlier comment, it's important to specify that this refers to the AArch64 variant of entry_point_info. You might even want to call this XFERLIST_EXEC_EP_INFO64.

source/transfer_list.rst
danh-arm: I'm glad you've managed to make the single EP for TE work.

Now there is only a single ep_info in this field, I tend to agree with Simon that saying the size of this field is data_size implies that this is of variable size when in fact it will always be sizeof(ep_info). So perhaps better to say that instead since it will become incorrect if this TE is ever extended.

I notice the same issue exists in the pp_addr and spmc_man fields of previous TEs. It would be nice If you can fix those as well (in a separate commit) but not critical for this PR.

danh-arm (changes requested):
Hopefully, nearly there!

source/transfer_list.rst
harrisonmutai-arm: Thank you, @danh-arm, for the feedback. I’ve created a separate pull request to address the same issue for pp_addr. Since spmc_man is a device tree, using data_size might be more suitable in this context.

https://github.com/FirmwareHandoff/firmware_handoff/pull/35

danh-arm (approved):

harrisonmutai-arm:
Hello @sjg20, are you able to take another look at this please? If you don't have any objections we can look to get it merged.

manish-pandey-arm:
@sjg20 If you are happy with this PR, could you please provide your votes?

manish-pandey-arm (approved):

manish-pandey-arm:
This PR is inactive for quite sometime, if there is no further comments can we please go ahead and merge?
@sjg20 @raymo200915

jmarinho (approved):
Hi,
The PR changes looks good to me.
After reviewing the discussion thread, my interpretation is that the main feedback was addressed:
- The TE has a single ep_info

Are we all in agreement to merge this PR?

raymo200915 (approved):
'multiple TEs with same tag ID' sounds good to me.

sjg20:
> @sjg20 If you are happy with this PR, could you please provide your votes?

See my comment: "What is this? It needs to be defined here, right? Other than that this looks good to me."

This code-review system is truly awful.

manish-pandey-arm:
> This code-review system is truly awful.
Totally agree with you!
> See my comment: "What is this? It needs to be defined here, right? Other than that this looks good to me."
Unable to find this comment (thanks to review system). But if your comments were about following then it has already been resolved (its outdated now)

     **- The size of SPMC manifest in bytes.

   * - ep_info
     - data_size**

sjg20:
> > This code-review system is truly awful.
> 
> Totally agree with you!
> 
> > See my comment: "What is this? It needs to be defined here, right? Other than that this looks good to me."
> 
> Unable to find this comment (thanks to review system). But if your comments were about following then it has already been resolved (its outdated now)
> 
> ```
>  **- The size of SPMC manifest in bytes.
> ```
> 
> * * ep_info
>   * data_size**

Is says: Holds a single `entry_point_info` structure.

So I believe that needs to be defined in this commit

harrisonmutai-arm:
> > > This code-review system is truly awful.
> > 
> > 
> > Totally agree with you!
> > > See my comment: "What is this? It needs to be defined here, right? Other than that this looks good to me."
> > 
> > 
> > Unable to find this comment (thanks to review system). But if your comments were about following then it has already been resolved (its outdated now)
> > ```
> >  **- The size of SPMC manifest in bytes.
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > 
> > * * ep_info
> >   * data_size**
> 
> Is says: Holds a single `entry_point_info` structure.
> 
> So I believe that needs to be defined in this commit

I believe the definition is covered here: https://github.com/FirmwareHandoff/firmware_handoff/pull/31/files#diff-75d1d8518f52351181570a42b6cee154a72b2eb15b3771b969f182b2e537b41eR908-R915.

What other details are you interested in, outside of what is already mentioned?

sjg20:
> > > > This code-review system is truly awful.
> > > 
> > > 
> > > Totally agree with you!
> > > > See my comment: "What is this? It needs to be defined here, right? Other than that this looks good to me."
> > > 
> > > 
> > > Unable to find this comment (thanks to review system). But if your comments were about following then it has already been resolved (its outdated now)
> > > ```
> > >  **- The size of SPMC manifest in bytes.
> > > ```
> > > 
> > > 
> > >     
> > >       
> > >     
> > > 
> > >       
> > >     
> > > 
> > >     
> > >   
> > > 
> > > * * ep_info
> > >   * data_size**
> > 
> > 
> > Is says: Holds a single `entry_point_info` structure.
> > So I believe that needs to be defined in this commit
> 
> I believe the definition is covered here: https://github.com/FirmwareHandoff/firmware_handoff/pull/31/files#diff-75d1d8518f52351181570a42b6cee154a72b2eb15b3771b969f182b2e537b41eR908-R915.
> 
> What other details are you interested in, outside of what is already mentioned?

So you are referring to another doc to describe the binary format? What if it changes?

I cannot see what the [TF_BL31] _ link refers to. Is it a page of a document, or a website, or something else?

manish-pandey-arm:
> > > > > This code-review system is truly awful.
> > > > 
> > > > 
> > > > Totally agree with you!
> > > > > See my comment: "What is this? It needs to be defined here, right? Other than that this looks good to me."
> > > > 
> > > > 
> > > > Unable to find this comment (thanks to review system). But if your comments were about following then it has already been resolved (its outdated now)
> > > > ```
> > > >  **- The size of SPMC manifest in bytes.
> > > > ```
> > > > 
> > > > 
> > > >     
> > > >       
> > > >     
> > > > 
> > > >       
> > > >     
> > > > 
> > > >     
> > > >   
> > > > 
> > > > * * ep_info
> > > >   * data_size**
> > > 
> > > 
> > > Is says: Holds a single `entry_point_info` structure.
> > > So I believe that needs to be defined in this commit
> > 
> > 
> > I believe the definition is covered here: https://github.com/FirmwareHandoff/firmware_handoff/pull/31/files#diff-75d1d8518f52351181570a42b6cee154a72b2eb15b3771b969f182b2e537b41eR908-R915.
> > What other details are you interested in, outside of what is already mentioned?
> 
> So you are referring to another doc to describe the binary format? What if it changes?
> 
> I cannot see what the [TF_BL31] _ link refers to. Is it a page of a document, or a website, or something else?

This is part of TF-A documentation and this structure is designed to support compatibility and independent evolution as it is versioned. This structure is constant from start of TF-A project and it is unlikely that it would change in future. Considering the introduced TE is in TF-A specific range, TF-A documentation is right place to provide the details.
Please refer to section "5.4.1.6.1.3. Data structures used in the BL31 cold boot interface"  of https://trustedfirmware-a.readthedocs.io/en/latest/

danh-arm:
This PR seems to be going around in circles. Outstanding comments were addressed in previous reworks. I intend to merge this in the next day unless there's feedback of material importance (or someone gets there before me).

source/references.rst
```diff
@@ -14,3 +14,5 @@
 .. [TFAFFAMB] `TF-A Secure Partition Manager: FF-A manifest binding to device tree <https://trustedfirmware-a.readthedocs.io/en/latest/components/ffa-manifest-binding.html>`
 
 .. [TCG_EFI] `https://trustedcomputinggroup.org/resource/tcg-efi-protocol-specification`.
+
+.. [TF_BL31] `Data structures used in the BL31 cold boot interface <https://trustedfirmware-a.readthedocs.io/en/latest/design/firmware-design.html#data-structures-used-in-the-bl31-cold-boot-interface>`
```
danh-arm, line 18: Sorry, just spotted that the hyperlink syntax in these references are wrong (needs a trailing __). Can you fix this? See https://github.com/FirmwareHandoff/firmware_handoff/pull/39 that fixes the other links.

danh-arm (approved):

sjg20:
> This PR seems to be going around in circles. Outstanding comments were addressed in previous reworks. I intend to merge this in the next day unless there's feedback of material importance (or someone gets there before me).

Unfortunately I don't agree with pointing to external documentation (via a hyperlink) to describe a structure. This spec should standard alone.

So for my approval, I would like to see the C struct spelled out here.

That said, everything else looks fine (as I said before)

danh-arm (approved):

source/transfer_list.rst
apalos: @raymo200915 recently pointed me to this thread. I think there are very valid concerns raised for having multiple entries with the same ID. AFAICT you are covering a very specific use case here, were everything is consumed by TF-A. But in the future people might want to add multiple DT entries for example -- Imagine a single bootloader that supports multiple SoCs and the only difference is the DT.
In that case blindly adding multiple entries is not enough. We need to be able to uniquely identify an entry.

I'll go through the comments again, but I think this needs some fixing

raymo200915: @apalos 
By looking closer to this issue, I think the user has two options when they need multiple TEs for same or similar purpose.

If the caller of each boot stage has the knowledge on which entry it should take (e.g. the current scenario of TF-A), then,
Optioin 1: Create multiple TEs with same tag ID, then get the array of entries by inputing the tag ID.

Otherwise,
Option 2: Define new entries as non-standard ones, so that it has a custom tag ID in the range reserved for non-standard TEs - Adding non-standard TE doesn't affect the spec changes.
This option can be used for the scenario of multiple DTs.

PR #36, opened 2024-02-09
Add RW memory layout TE

harrisonmutai-arm:
XFERLIST_RW_MEM_LAYOUT64 is used to specify an extent of read-write memory that a client can use. Note, in future if other memory types need representation seperate TE's should be defined (left in the layout description as a point of discussion).

### DIFF
--- a/source/transfer_list.rst
+++ b/source/transfer_list.rst
@@ -902,5 +902,49 @@ This manifest contains the SPMC attribute node consumed by the SPMD
      - hdr_size
      - Holds a SPMC manifest image in DT format.
 
+**Read-Write Memory Layout Entry Layout (XFERLIST_RW_MEM_LAYOUT64)**
+
+This entry type holds a structure that describes the layout of a read-write
+memory region. TF-A defines this structure and uses it to convey to BL2 the
+extent of memory it has available to perform read-write operations on. BL2 maps
+the memory described by the layout into its memory map during platform setup. If
+other memory types are required (i.e.  read-only memory) separate TE's should be
+defined.
+
+.. _tab_rw_mem_layout:
+.. list-table:: Layout for a RW memory layout entry
+   :widths: 2 5 5 6
+
+   * - Field
+     - Size (bytes)
+     - Offset (bytes)
+     - Description
+
+   * - tag_id
+     - 0x3
+     - 0x0
+     - The tag_id field must be set to **104**.
+
+   * - hdr_size
+     - 0x1
+     - 0x3
+     - |hdr_size_desc|
+
+   * - data_size
+     - 0x4
+     - 0x4
+     - The size of the layout in bytes.
+
+   * - addr
+     - 8
+     - hdr_size
+     - The base address of the memory region.
+
+   * - size
+     - 8
+     - hdr_size + 8
+     - The size of the memory region.
+
+
 .. |hdr_size_desc| replace:: The size of this entry header in bytes must be set to **8**.
 .. |current_version| replace:: `0x1`
--- a/source/transfer_list.rst
+++ b/source/transfer_list.rst
@@ -910,7 +910,7 @@ memory region.
 For example, TF-A uses it to convey to BL2 the extent of memory it has available
 to perform read-write operations on. BL2 maps the memory described by the layout
 into its memory map during platform setup. If other memory types are required
-(i.e.  read-only memory) separate TE's should be defined.
+(i.e. read-only memory) separate TE's should be defined.
 
 .. _tab_rw_mem_layout:
 .. list-table:: Layout for a RW memory layout entry
--- a/source/transfer_list.rst
+++ b/source/transfer_list.rst
@@ -937,13 +937,13 @@ into its memory map during platform setup. If other memory types are required
      - The size of the layout in bytes.
 
    * - addr
-     - 8
+     - 0x8
      - hdr_size
      - The base address of the memory region.
 
    * - size
-     - 8
-     - hdr_size + 8
+     - 0x8
+     - hdr_size + 0x8
      - The size of the memory region.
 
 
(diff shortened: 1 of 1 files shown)


source/transfer_list.rst
```diff
@@ -820,6 +821,48 @@ and reserved for use as a TPM Command Response Buffer interface.
...
 
 
+.. _rw_mem_layout:
+
+Read-Write Memory Layout Entry Layout (XFERLIST_RW_MEM_LAYOUT)
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+This type defines the layout of a region of read-write memory that a client can
+allocate into its memory map. If other memory types are required (i.e.
+read-only memory), separate TE's should be defined.
```
jwerner-chromium, line 831: This sounds very specific to a single project (TF-A?), so it should probably go in the TE section for that. At least I wouldn't know from this description what to do with this entry if I encountered it in my firmware.

source/transfer_list.rst
```diff
@@ -820,6 +821,48 @@ and reserved for use as a TPM Command Response Buffer interface.
...
+     - 0x3
+     - |hdr_size_desc|
+
+   * - data_size
+     - 0x4
+     - 0x4
+     - The size of the layout in bytes.
+
+   * - addr
+     - `sizeof(uintptr_t)`
```
jwerner-chromium, line 858: The TL is architecture-independent so you should avoid references to machine register sizes and use exact widths wherever possible.

source/transfer_list.rst
harrisonmutai-arm: You're probably right. The structure this TE represents is fairly simple:
``` c
typedef struct meminfo {
	uintptr_t total_base;
	size_t total_size;
} meminfo_t;
```
This to me seems adaptable to other contexts, although i'm yet to find any concrete examples. In any case, I've moved it to TF-A's namespace. If a more general use case materialises then we can consider adding it back.

source/transfer_list.rst
```diff
@@ -902,5 +902,49 @@ This manifest contains the SPMC attribute node consumed by the SPMD
...
      - Holds a SPMC manifest image in DT format.
 
+**Read-Write Memory Layout Entry Layout (XFERLIST_RW_MEM_LAYOUT64)**
+
+This entry type holds a structure that describes the layout of a read-write
+memory region. TF-A defines this structure and uses it to convey to BL2 the
+extent of memory it has available to perform read-write operations on. BL2 maps
+the memory described by the layout into its memory map during platform setup. If
+other memory types are required (i.e.  read-only memory) separate TE's should be
+defined.
```
sjg20, line 912: Would it be useful to generalise this to include any 'next phase'??

sjg20 (changes requested):

source/transfer_list.rst
harrisonmutai-arm: What do you mean? I refer to its usage in BL2 here only as a concrete example, but it very well could be used in other places.

source/transfer_list.rst
```diff
@@ -902,5 +902,49 @@ This manifest contains the SPMC attribute node consumed by the SPMD
      - hdr_size
      - Holds a SPMC manifest image in DT format.
 
+**Read-Write Memory Layout Entry Layout (XFERLIST_RW_MEM_LAYOUT64)**
+
+This entry type holds a structure that describes the layout of a read-write
+memory region. TF-A defines this structure and uses it to convey to BL2 the
```
sjg20, line 908: Should the sentences after the first be written as an example? It seems that this information could be used elsewhere

source/transfer_list.rst
sjg20: OK, then please put 'For example, ...' in a separate paragraph so it is clear

sjg20 (changes requested):

source/transfer_list.rst
```diff
@@ -902,5 +902,50 @@ This manifest contains the SPMC attribute node consumed by the SPMD
...
+     - 0x3
+     - |hdr_size_desc|
+
+   * - data_size
+     - 0x4
+     - 0x4
+     - The size of the layout in bytes.
+
+   * - addr
+     - 8
```
danh-arm, line 940: The de-facto standard is to use hex values in these tables, so 0x8.

source/transfer_list.rst
```diff
@@ -902,5 +902,50 @@ This manifest contains the SPMC attribute node consumed by the SPMD
...
+     - hdr_size
+     - The base address of the memory region.
+
+   * - size
+     - 8
+     - hdr_size + 8
```
danh-arm, line 946: Same again. use 0x8.

danh-arm (changes requested):
Generally looks good. Just minor comments.

danh-arm (approved):

source/transfer_list.rst
```diff
@@ -902,5 +902,50 @@ This manifest contains the SPMC attribute node consumed by the SPMD
...
 
+**Read-Write Memory Layout Entry Layout (XFERLIST_RW_MEM_LAYOUT64)**
+
+This entry type holds a structure that describes the layout of a read-write
+memory region.
+
+For example, TF-A uses it to convey to BL2 the extent of memory it has available
+to perform read-write operations on. BL2 maps the memory described by the layout
+into its memory map during platform setup. If other memory types are required
+(i.e.  read-only memory) separate TE's should be defined.
```
sjg20, line 913: nit: change to single space before 'read-only'

sjg20 (dismissed):
LGTM (with nit)

danh-arm (approved):

manish-pandey-arm (approved):

PR #37, opened 2024-02-09
Add Trusted Boot Firmware Configuration entry

harrisonmutai-arm:
This entry type holds the flattened device tree used to configure trusted firmware, it is specific the to the Trusted Firmware-A project.

### DIFF
--- a/source/transfer_list.rst
+++ b/source/transfer_list.rst
@@ -950,6 +950,43 @@ software running in Secure, Non-Secure, or Realm modes.
      - hdr_size
      - Holds a single `entry_point_info` structure.
 
+**DT Trusted Boot Firmware Configuration entry layout
+(XFERLIST_TB_FW_CONFIG)**
+
+The Trusted Boot Firmware Configuration (TB_FW_CONFIG) is a dynamic
+configuration file used in TF-A to configure properties relating to trusted
+firmware (i.e. IO policies, and Mbed TLS heap info) [TB_FW_CONFIG]_. This TE
+contains in its data section the configuration file in DT format [DT]_.
+
+.. _tab_tb_fw_config:
+.. list-table:: Trusted Boot Firmware Configuration entry type layout
+   :widths: 2 2 2 8
+
+   * - Field
+     - Size (bytes)
+     - Offset (bytes)
+     - Description
+
+   * - tag_id
+     - 0x3
+     - 0x0
+     - The tag_id field must be set to **0x103**.
+
+   * - hdr_size
+     - 0x1
+     - 0x3
+     - |hdr_size_desc|
+
+   * - data_size
+     - 0x4
+     - 0x4
+     - The size of the configuration file in bytes.
+
+   * - tb_fw_config
+     - data_size
+     - hdr_size
+     - Holds a Trusted Boot Firmware Configuration file in DT format.
+
 **Read-Write Memory Layout Entry Layout (XFERLIST_RW_MEM_LAYOUT64)**
 
 This entry type holds a structure that describes the layout of a read-write
--- a/source/references.rst
+++ b/source/references.rst
@@ -28,3 +28,5 @@ References
 .. [TCG_EFI] `Trusted Computing Group EFI Protocol Specification <https://trustedcomputinggroup.org/resource/tcg-efi-protocol-specification>`__
 
 .. [TF_BL31] `Data structures used in the BL31 cold boot interface <https://trustedfirmware-a.readthedocs.io/en/latest/design/firmware-design.html#data-structures-used-in-the-bl31-cold-boot-interface>`__
+
+.. [TB_FW_CONFIG] `Trusted Boot Firmware Configuration bindings <https://trustedfirmware-a.readthedocs.io/en/latest/components/fconf/tb_fw_bindings.html>`__


source/transfer_list.rst
```diff
@@ -902,5 +902,42 @@ This manifest contains the SPMC attribute node consumed by the SPMD
      - hdr_size
      - Holds a SPMC manifest image in DT format.
 
+**DT Trusted Boot Firmware Configuration entry layout
```
sjg20, line 905: Would it be possible to put this information into the control dtb?

sjg20 (changes requested):

source/transfer_list.rst
harrisonmutai-arm: No I don't think it would, the information in this dtb ( i.e. chain-of-trust, and mbedtls settings) is only intended for consumption by the trusted bootloader. Putting this information in the hardware dtb would require a client to strip this information prior to handover to later stages.

sjg20: Why is that? Where are the bindings for these nodes/properties?

harrisonmutai-arm: The nodes/properties are used specifically for the configuration of firmware. They aren't intended for use by any other components, and are unrelated to hardware descriptions. In some cases, the nodes are meant to be secret. Here is an example of the chain-of-trust bindings: https://review.trustedfirmware.org/plugins/gitiles/TF-A/trusted-firmware-a/+/refs/heads/master/docs/components/cot-binding.rst.

sjg20: > The nodes/properties are used specifically for the configuration of firmware. They aren't intended for use by any other components, and are unrelated to hardware descriptions. In some cases, the nodes are meant to be secret. Here is an example of the chain-of-trust bindings: https://review.trustedfirmware.org/plugins/gitiles/TF-A/trusted-firmware-a/+/refs/heads/master/docs/components/cot-binding.rst.

Thank you for the info. The binding looks like a good start but shouldn't it be in yaml format? How is this validated? I don't see mention of secret nodes. We don't have a way (at present) of specifying when nodes should be deleted before the next boot phase is invoked.

I see that the widevine binding is in the 'other' DT. This will get very confusing.

In general this seems to be a can of worms. We should have one DT schema and work to get it upstream

sjg20: > Each FW image may have its own FW_CONFIG contain data that only of relevance to its receiver and not subsequent components. This data is orthogonal to the HW device tree and device tree format is simply used as a convenient format for this data. This is all internal TF-A design stuff and doesn't need to be specified more widely.

This seems to open a distinction between the devicetree used by the system as a whole and the one used by TF-A. I have been pushing back against attempts to have U-Boot have its own DT. If TF-A has this, then we will need U-Boot to access two different DTs, won't we?

Why have a separate one? I mean, what is the benefit?

manish-pandey-arm: > This seems to open a distinction between the devicetree used by the system as a whole and the one used by TF-A. 

TF-A consumes both system device tree and this special device tree
This TE is for Trusted Boot Configuration data which is consumed by BL2 (image loader) phase of TF-A. To load images with chain of trust we need image UUIDs and for some images (like Secure partitions) we also need load addresses. Considering we have a lot of images (32 standard images + Any number of SP's) we use device tree as a configuration file. 

> If TF-A has this, then we will need U-Boot to access two different DTs, won't we?

sjg20: Thanks for the responses.

I would like to see the lifetime of this information defined...e.g. you suggest that some pieces must not be passed onto the OS. Which ones and why not? It should be written out explicitly.

Also if we want to use a single DT, is that possible / allowed?

I also wish to note that you do not intend (I believe) to get these bindings accepted in dt-schema.

I suspect you plan to merge this anyway, but at least you have my objection noted. This will be the second PR merged over my objections.

danh-arm (dismissed):
Please resolve conflicts and squash the commits.

danh-arm (approved):

danh-arm (approved):

source/transfer_list.rst
harrisonmutai-arm: Hi Simon,

I've created a binding document within the TF-A documentation that outlines the usage of the TB_FW_CONFIG format. [This document is referenced in the specification and contains all the essential information](https://trustedfirmware-a.readthedocs.io/en/latest/components/fconf/tb_fw_bindings.html). Let us know if you're satisfied with this and we can look to get the change merged.

sjg20: Hi again,

Thank you for the binding. It helps a lot!

The properties should have hyphens, not underscores. It isn't clear to me why UUIDs are in there. The binding documentation is mostly not enough (for me) to figure out what it all means. For one thing, I don't see any mention of nodes.

Anyway the purpose of this PR is not to agree the binding. It just seems clear to me that the binding as is would need changes for dt-schema. Other projects have gone to enormous efforts to get bindings agreed (e.g. U-Boot). Could TF-A put in some effort to?

Could you clarify the lifetime of this information, as I requested?

Overall I believe it would be better to get an official binding agreed.

manish-pandey-arm: > Hi again,
> 
> Thank you for the binding. It helps a lot!
> 
Thanks for your review!
> The properties should have hyphens, not underscores. It isn't clear to me why UUIDs are in there. The binding documentation is mostly not enough (for me) to figure out what it all means. For one thing, I don't see any mention of nodes.
> 
We currently are looking at decoupling individual properties in this dt to be created as seperate TE (https://github.com/FirmwareHandoff/firmware_handoff/pull/45 for mbetls_heap). Also, doing some experiments to see if we can live without UUIDs
> Anyway the purpose of this PR is not to agree the binding. It just seems clear to me that the binding as is would need changes for dt-schema. Other projects have gone to enormous efforts to get bindings agreed (e.g. U-Boot). Could TF-A put in some effort to?
> 
One of the purpose of Firmware handoff was to stop usage of device trees as Configuration files and we certainly will reduce dt usage in TF-A. Our plan is to reduce number of dt properties present in TF-A and for those which we can't remove, will put efforts to make the bindings in dt-schema.
> Could you clarify the lifetime of this information, as I requested?
> 
By lifetime if you mean the stage which would consume this information then it is BL2.
> Overall I believe it would be better to get an official binding agreed.
Ack

sjg20 (commented):
I have a request for a review, but I believe I have done this.

The right thing to do here is to add a binding for the DT notes, but I think there is a larger disconnect, as on the call this morning I believe I heard Dan mention 'hardware devicetree', suggesting that there are multiple.

IMO there should not be multiple, or at least nothing should break if nodes not intended for the OS make their way to the OS

harrisonmutai-arm:
Thanks all for the feedback. We are looking at a different approach that moves information from the TB FW config into transfer entries. As a result, closing this pull request for now.

PR #38, opened 2024-02-23
add DICE TL entries

thomas-fossati:
TL entries used for [Open-DICE](https://pigweed.googlesource.com/open-dice/+/refs/heads/main/docs/specification.md), including CDIs (for attestation and optional sealing) and certificates' chain.

### DIFF
--- a/source/references.rst
+++ b/source/references.rst
@@ -27,4 +27,4 @@ References
 
 .. [TCG_EFI] `Trusted Computing Group EFI Protocol Specification <https://trustedcomputinggroup.org/resource/tcg-efi-protocol-specification>`__
 
-.. [TF_BL31] `Data structures used in the BL31 cold boot interface <https://trustedfirmware-a.readthedocs.io/en/latest/design/firmware-design.html#data-structures-used-in-the-bl31-cold-boot-interface>`
+.. [TF_BL31] `Data structures used in the BL31 cold boot interface <https://trustedfirmware-a.readthedocs.io/en/latest/design/firmware-design.html#data-structures-used-in-the-bl31-cold-boot-interface>`__


source/transfer_list.rst
```diff
@@ -902,5 +903,128 @@ This manifest contains the SPMC attribute node consumed by the SPMD
...
+
+The following entry types are currently defined:
+
+**CDI for attestation (XFERLIST_DICE_CDI_ATTEST)**
+
+This entry type holds the CDI described in [DICE]_.
+In [Open-DICE]_ it is referred to as "Attestation CDI".
+
+The CDI computed by the Sender's boot stage SHALL be added to the TL.
+The CDI received from the previous boot stage (i.e., DICE layer) MUST NOT be forwarded.
```
apalos, line 923: I know it's described in the TCG docs, but shouldn't we also require the CDI user to clean the memory ?

thomas-fossati: It certainly doesn't hurt, I agree.

source/transfer_list.rst
```diff
@@ -902,5 +903,128 @@ This manifest contains the SPMC attribute node consumed by the SPMD
...
+
+.. _tab_dice_cdi_attest:
+.. list-table:: DICE attestation CDI
```
apalos, line 926: So, I am agree with adding the openDICE profile. Would it make sense to correlate this one to the TCG Alias ID?
The attestation CDI seems similar to the Alias one described in the spec.

source/transfer_list.rst
thomas-fossati: See if [0bb6617](https://github.com/FirmwareHandoff/firmware_handoff/pull/38/commits/0bb6617400214acf8a2fef3b81a49e074feac46d) works for you.

source/transfer_list.rst
```diff
@@ -908,9 +908,15 @@ This manifest contains the SPMC attribute node consumed by the SPMD
 Entries related to DICE
 ^^^^^^^^^^^^^^^^^^^^^^^
 
-These entries hold [DICE]_ and [Open-DICE]_ information elements that needs to be transferred from one boot stage to the next.
+These entries hold [DICE]_ and [Open-DICE]_ information elements that needs to be transferred from one boot stage to the next in a *layered* DICE architecture, i.e., one that does not emply the DICE Protection Environment (DPE).
```
apalos, line 911: emply == imply?

source/transfer_list.rst
```diff
@@ -908,9 +908,15 @@ This manifest contains the SPMC attribute node consumed by the SPMD
...
 
-This includes the CDIs used for attestation (and optional sealing), as well as the accumulated certificate chain.
+This includes the Compound Device Identifier (CDI) used for attestation (and optional sealing), as well as the accumulated certificate chain.
+
+.. note::
+  As described in [DICE]_, CDIs are very sensitive data and are as long-lived as the layers they represent.
+  Each layer is therefore responsible for the proper handling of the CDIs it owns, as well as the derived cryptographic key material.
+  In particular, any given layer is expected to use and erase the CDI it received.
+  While the CDI is used, the layer needs the ability to prevent access to the CDI and disclosure of the value.
```
apalos, line 919: yes that's fine

source/transfer_list.rst
thomas-fossati: > So, I agree with adding the openDICE profile. Would it make sense to correlate this one to the TCG Alias ID? The attestation CDI seems similar to the Alias one described in the spec.

The Alias Key described in [Section 8 of DICE LA](https://trustedcomputinggroup.org/wp-content/uploads/DICE-Layering-Architecture-r19_pub.pdf#page=18) is the leaf key in the DICE cert chain, i.e., the one that is obtained from the last CDI in the TCB's components "stack".  This key is computed (also) using the TCI (i.e., at a minimum, the digest of the firmware code) of the last layer.  So it's not as stable as Open-DICE [sealing key](https://pigweed.googlesource.com/open-dice/+/0b5044098bf9b40128927d675dea4ec1fb75c510/docs/specification.md#cdi-values), which is computed from authority data, mode decision, and hidden inputs, i.e., all stable inputs.

So, structurally TCG Alias Key and Open-DICE sealing keys are quite different objects.  Also semantically: an Alias Key is the one that signs attestation challenges, while a sealing key is a key that encrypts some data (and binds that encryption to a specific state of the device/program).

source/transfer_list.rst
thomas-fossati: Good catch, thanks!  I meant "employ."

source/transfer_list.rst
```diff
@@ -902,5 +903,134 @@ This manifest contains the SPMC attribute node consumed by the SPMD
...
+     - 0x3
+     - |hdr_size_desc|
+
+   * - data_size
+     - 0x4
+     - 0x4
+     - The size (in bytes) of the CDI
+
+   * - cdi_attest
+     - 0x8
```
jmarinho, line 956: is this "data_size" instead of 8?

thomas-fossati: yes, thanks!  I got them all wrong.

thomas-fossati: See [b9f4bc5](https://github.com/FirmwareHandoff/firmware_handoff/pull/38/commits/b9f4bc567a548364fedfbe9b1a807022655352ac)

source/transfer_list.rst
apalos: I wasn't referring to the sealing key, but it wasn't entirely clear. 
I meant the alias ID and the attestation one described in openDICE seems similar

thomas-fossati: > I wasn't referring to the sealing key, but it wasn't entirely clear

ah, sorry I misunderstood.

> I meant the alias ID and the attestation one described in openDICE seem similar

In TCG, the alias is the leaf cert, so I guess by "alias ID" you mean the CDI associated with the last TCB layer?  If so, that's a very specific CDI.  Instead, the "attestation CDI" in Open-DICE is a generic name for *any* CDI computed by DICE.  It's the equivalent of TCG's CDI - without qualifications.

apalos: The 'alias ID' parlance comes from https://www.microsoft.com/en-us/research/project/dice-device-identifier-composition-engine/. Looking at it, it seems to include measurements of various layers not only the last one. IOW it seems to measure all subsequent layers

thomas-fossati: > The 'alias ID' parlance comes from https://www.microsoft.com/en-us/research/project/dice-device-identifier-composition-engine/. Looking at it, it seems to include measurements of various layers not only the last one. IOW it seems to measure all subsequent layers

OK, we were looking at different specs :-)

Yes, in DICE/RiOT sense you are right.  But a few years ago (about 8 I think) MSFT gave TCG change control over DICE.  And TCG has made some adjustments to the vocabulary, including narrowing the concept of "alias" to the outer layer only.

It is very confusing that MSFT hasn't clearly marked the RiOT spec as stale :-(

source/transfer_list.rst
```diff
@@ -902,5 +903,134 @@ This manifest contains the SPMC attribute node consumed by the SPMD
...
+     - The size (in bytes) of the sealing CDI
+
+   * - cdi_seal
+     - data_size
+     - hdr_size
+     - Holds the CDI used for sealing
+
+**Certificate chain (XFERLIST_DICE_CERT_CHAIN)**
+
+This entry type holds the DICE certificate chain.
```
danh-arm, line 998: What precisely is this? I guess a set of concatenated certificates with no metadata or padding between them? I wonder if it would be better for this to be a single certificate, with multiple instances representing the entire chain? That would make it easier for the chain to be extended as it's passed between components, or if only part of the chain needs to be passed around.
Also, what kind of certificates are represented here? The OpenDICE spec refers to UDS and CDI certificates so I guess you mean those? I presume there are also attestation and sealing certificates but the spec doesn't seem to specify those.

danh-arm (changes requested):

source/references.rst
```diff
@@ -26,3 +26,6 @@ References
 .. [TFAFFAMB] `TF-A Secure Partition Manager: FF-A manifest binding to device tree <https://trustedfirmware-a.readthedocs.io/en/latest/components/ffa-manifest-binding.html>`
 
 .. [TCG_EFI] `https://trustedcomputinggroup.org/resource/tcg-efi-protocol-specification`.
+
+.. [DICE] `https://trustedcomputinggroup.org/wp-content/uploads/DICE-Layering-Architecture-r19_pub.pdf`.
+.. [Open-DICE] `https://pigweed.googlesource.com/open-dice/+/refs/heads/main/docs/specification.md`.
```
danh-arm, line 31: Please fix the hyperlink syntax in these references. See https://github.com/FirmwareHandoff/firmware_handoff/pull/39 that fixes the other links.

source/transfer_list.rst
thomas-fossati: > What precisely is this? I guess a set of concatenated certificates with no metadata or padding between them?

Yes, that was the idea.

> I wonder if it would be better for this to be a single certificate, with multiple instances representing the entire chain?

Good point.

> That would make it easier for the chain to be extended as it's passed between components, or if only part of the chain needs to be passed around.

I am not sure I can see a pruning use case, but your simplification argument is a good one.

> Also, what kind of certificates are represented here? The OpenDICE spec refers to UDS and CDI certificates so I guess you mean those?

Yes.

> I presume there are also attestation and sealing certificates but the spec doesn't seem to specify those.

The leaf cert in the chain is the "attestation" cert, i.e., the one that is supposed to sign challenges from the user.  The UDS cert is the "Device ID", while the rest are just intermediate CA certs.

There are no sealing certs, only sealing CDIs that can be used to derive sealing keys.

source/transfer_list.rst
```diff
@@ -902,5 +903,134 @@ This manifest contains the SPMC attribute node consumed by the SPMD
...
+     - 0x1
+     - 0x3
+     - |hdr_size_desc|
+
+   * - data_size
+     - 0x4
+     - 0x4
+     - The size (in bytes) of the CDI
+
+   * - cdi_attest
```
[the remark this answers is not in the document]
thomas-fossati: Would a "profile identifier" in the header do the trick?

source/references.rst
thomas-fossati: done in [fd0e476](https://github.com/FirmwareHandoff/firmware_handoff/pull/38/commits/fd0e476b924f77c34e6694569b2f369a4cd97542)

source/transfer_list.rst
```diff
@@ -902,5 +903,134 @@ This manifest contains the SPMC attribute node consumed by the SPMD
      - hdr_size
      - Holds a SPMC manifest image in DT format.
 
+.. _dice_entries:
+
+Entries related to DICE
+^^^^^^^^^^^^^^^^^^^^^^^
+
+These entries hold [DICE]_ and [Open-DICE]_ information elements that needs to be transferred from one boot stage to the next in a *layered* DICE architecture, i.e., one that does not employ the DICE Protection Environment (DPE).
```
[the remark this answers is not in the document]
thomas-fossati: done in [c4dc061](https://github.com/FirmwareHandoff/firmware_handoff/pull/38/commits/c4dc0617db86bc540d431cf2ba5e6f40db36c527)

source/transfer_list.rst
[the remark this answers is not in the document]
danh-arm: > Would a "profile identifier" in the header do the trick?
Maybe, if there was a doc that mapped each supported "profile identifier" to a binary data format.

source/transfer_list.rst
danh-arm: OK, so can we expect another change to make this an OpenDICE certificate TE (instead of a certificate chain)?

> The leaf cert in the chain is the "attestation" cert

OK, so all certs are either OpenDICE UDS or CDI certs. Would be good to clarify this in the text.

source/transfer_list.rst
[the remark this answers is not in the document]
thomas-fossati: > > Would a "profile identifier" in the header do the trick?
> 
> Maybe, if there was a doc that mapped each supported "profile identifier" to a binary data format.

In [Open-DICE v2.5](https://pigweed.googlesource.com/open-dice/+/refs/heads/main/docs/specification.md), `OpenDiceInput` has a `profileName` claim for that purpose.

At least Android's Open DICE specialisations have a document describing profile-specific instantiation choices, which are mapped to a bunch of [profile identifier strings](https://pigweed.googlesource.com/open-dice/+/refs/heads/main/docs/android.md#Versions).

For TCG DICE the situation is much more foggy.  All of the baseline (i.e., non-DPE) specs allow ample freedom for implementation choices, but they don't have an explicit protocol indicator for specifying the specific instantiations.  The upcoming DPE spec does, but only concerning the DPE itself.

In summary, ISTM that a sensible way forward could be to drop TCG support and scope the TLs in this PR to Open DICE only, adding a "profile id" field to all the newly defined headers.

WDYT?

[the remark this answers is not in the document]
danh-arm: Actually, on reading OpenDICE some more, I don't think it's important for receivers to understand the binary format of `cdi_attest`. They blindly use this as input to the KDF when generating the next layer's `cdi_attest`.

Also, `profileName` is part of the specified OpenDICE CDI certificate so I don't think there's anything additional to specify here.

I agree that TEs in this PR should be specific to OpenDICE. However, I'm reluctant to invent a "profile id" field with specific instantiation values if TCG don't already do this. I don't think we're likely to get lots of (or even any) other (non-DPE) profiles we need to support with this, right?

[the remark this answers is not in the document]
thomas-fossati: > Actually, on reading OpenDICE some more, I don't think it's important for receivers to understand the binary format of `cdi_attest`. They blindly use this as input to the KDF when generating the next layer's `cdi_attest`.

Sorry, I wanted to reply specifically on that point but got lost in the profile rabbit hole...  As you say above, CDIs are just byte strings they don't need any special treatment.

> Also, `profileName` is part of the specified OpenDICE CDI certificate so I don't think there's anything additional to specify here.

For chaining to work all layers need to synchronise on the crypto primitives.  So, there must be an explicit way to communicate that information at handover points.  One option is to pass all the details in-band (hash, KDF, and digital signatures algorithms).  Another option is to use indirection via the profile concept.  What would be your choice?

> I agree that TEs in this PR should be specific to OpenDICE. However, I'm reluctant to invent a "profile id" field with specific instantiation values if TCG don't already do this.

If we only focus on OpenDICE we are not restricted in our choices.  We can deal with TCG DICE if and when needed, adding another TL if the ones already defined are not good enough.

> I don't think we're likely to get lots of (or even any) other (non-DPE) profiles we need to support with this, right?

I plan to define one such OpenDICE profile and use it, but effectively I don’t know of anyone else wanting to do the same, at least at this time.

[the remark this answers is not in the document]
danh-arm: > For chaining to work all layers need to synchronise on the crypto primitives...

OK, then it does sound important for the next layer to understand the previous layer's `profileName` (before any certificates are constructed). I think it's enough to pass this info in the TEs rather than all the CDI details.

> I plan to define one such OpenDICE profile and use it,..

Sounds good.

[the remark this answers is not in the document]
thomas-fossati: I gave it an initial stab at a89f7d4

PR #41, opened 2024-03-19
add TL entries for Open-DICE

thomas-fossati:
This PR is a rework of #38 which:

* narrows the scope to Open-DICE
* requires one TL for each cert in the chain
* adds an explicit profile_id field
* adds a registry for profile_id's
* populates the profile_id's registry with Android entries

### DIFF
--- a/source/transfer_list.rst
+++ b/source/transfer_list.rst
@@ -512,6 +512,7 @@ The following entry types are currently defined:
 - TPM event log entry: tag_id = 5 (:numref:`tpm_evlog_entry`).
 - TPM CRB base entry: tag_id = 6 (:numref:`tpm_crb_base_entry`).
 - Entries related to Trusted Firmware (:numref:`tf_entries`).
+- Entries related to Open DICE (:numref:`opendice_entries`).
 
 .. _void_entry:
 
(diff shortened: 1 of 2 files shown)


thomas-fossati:
cc @danh-arm @apalos @jmarinho

apalos (approved):

source/transfer_list.rst
```diff
@@ -994,6 +995,177 @@ into its memory map during platform setup. If other memory types are required
...
+     - 0x4
+     - The size (in bytes) of the CDI
+
+   * - profile_id
+     - 0x4
+     - 0x8
+     - the Open DICE profile identifier (See :numref:`opendice_profile_id` for the allowed values)
+
+   * - cdi_attest
+     - data_size - 0x4
```
jmarinho, line 1052: Are CDIs larger than 4 bytes?
If so the 0x4 is a typo?

The same question holds for lines 1093 and 1039.

source/transfer_list.rst
```diff
@@ -994,6 +995,177 @@ into its memory map during platform setup. If other memory types are required
...
+
+   * - Field
+     - Size (bytes)
+     - Offset (bytes)
+     - Description
+
+   * - tag_id
+     - 0x3
+     - 0x0
+     - The tag_id field must be set to **0x200**.
```
jmarinho, line 1034: I'd like to request @manish-pandey-arm to confirm these tag_ids are the best suited, or whether we should use lower values.

source/transfer_list.rst
thomas-fossati: > Are CDIs larger than 4 bytes? If so the 0x4 is a typo?
> 
> The same question holds for lines 1093 and 1039.

I am probably missing something fundamental about the notation conventions.  My reasoning was: the data (`data_size` long) comprises `profile_id` (fixed length, 4 bytes) and `cdi_attest` (variable length).  Therefore, `cdi_attest`'s length is `data_size - 0x4`.

jmarinho: Hi Thomas,

I think I was the one getting confused with the "-" and assumed the line to state something like "data_size, this should be  0x4".
All checks out. I'll resolve this converstation.

danh-arm (changes requested):
Please also add a commit message.

I'm OK with this PR in essence. My comments should be easy to address.

source/transfer_list.rst
manish-pandey-arm: Going back to https://github.com/FirmwareHandoff/firmware_handoff/pull/38 , i can't see the reasoning why this tag was moved from generic range to 0x200 ? It is a good candidate to stay in generic range.
As mentioned by Dan, i don't think TF-A going to consume more than 32 entries, so 0x120 should be a safe value (if we decide to keep in non-generic range)

thomas-fossati: Both options work for me:

1. Starting at 0x120
2. Using the generic range, in which case 8, 9 and 10 are the lowest three that are available - 7 is already taken by the heap carve-outs #45

Let me know what is your preference.

source/transfer_list.rst
```diff
@@ -994,6 +995,177 @@ into its memory map during platform setup. If other memory types are required
...
+
+Entries related to Open DICE
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+These entries contain [Open-DICE]_ information elements that must be transferred between boot stages in a DICE architecture where a DICE Protection Environment (DPE) is not used.
+
+This includes the Compound Device Identifier (CDI) used for attestation and the CDI used for sealing, as well as the accumulated certificate chain.
+
+.. note::
+  As described in [DICE]_, CDIs are very sensitive data and are as long-lived as the layers they represent.
```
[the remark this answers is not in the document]
thomas-fossati: done in [3147252](https://github.com/FirmwareHandoff/firmware_handoff/pull/41/commits/3147252ff159189a911a6605b61feee5fe84a3e8)

source/transfer_list.rst
```diff
@@ -994,6 +995,177 @@ into its memory map during platform setup. If other memory types are required
...
+     - data_size - 0x4
+     - 0xc
+     - Holds the CDI used for sealing
+
+**Certificate (XFERLIST_OPENDICE_CERT)**
+
+This entry type holds one certificate of the Open DICE certificate chain.
+
+No specific certificate format is mandated.
+Examples of certificate formats include X.509, CWT, and C509.
```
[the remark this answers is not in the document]
thomas-fossati: I will expand CWT and add a reference.

source/transfer_list.rst
```diff
@@ -994,6 +995,177 @@ into its memory map during platform setup. If other memory types are required
...
+   * - cdi_seal
+     - data_size - 0x4
+     - 0xc
+     - Holds the CDI used for sealing
+
+**Certificate (XFERLIST_OPENDICE_CERT)**
+
+This entry type holds one certificate of the Open DICE certificate chain.
+
+No specific certificate format is mandated.
```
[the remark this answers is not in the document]
thomas-fossati: I'll add a ref to the relevant section in the Open DICE spec.

source/transfer_list.rst
```diff
@@ -994,6 +995,177 @@ into its memory map during platform setup. If other memory types are required
...
+  While the CDI is used, the layer needs the ability to prevent access to the CDI and disclosure of the value.
+
+The following entry types are currently defined:
+
+**CDI for attestation (XFERLIST_OPENDICE_CDI_ATTEST)**
+
+This entry type holds the "Attestation CDI" described in [Open-DICE]_.
+
+The CDI computed by the Sender's boot stage SHALL be added to the TL.
+The CDI received from the previous boot stage (i.e., DICE layer) MUST NOT be forwarded.
```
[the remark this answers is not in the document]
thomas-fossati: > I think we need a comment here about what the receiver should do with the profile_id, e.g. use this to generate a CDI that is compatible with the previous layer's CDI as per the Android OpenDICE spec.

What is the behaviour in case $L_{i+1}$ does not implement a profile compatible with layer $L_i$?

source/transfer_list.rst
```diff
@@ -994,6 +995,177 @@ into its memory map during platform setup. If other memory types are required
...
+     - data_size - 0x4
+     - 0xc
+     - Holds the CDI used for attestation
+
+**CDI for sealing (XFERLIST_OPENDICE_CDI_SEAL)**
+
+This entry type holds the "Sealing CDI" described in [Open-DICE]_.
+
+The CDI computed by the Sender's boot stage SHALL be added to the TL.
+The CDI received from the previous boot stage (i.e., DICE layer) MUST NOT be forwarded.
```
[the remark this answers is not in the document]
thomas-fossati: OK, I'll add it here so that all requirements are kept together.

source/transfer_list.rst
```diff
@@ -994,6 +995,177 @@ into its memory map during platform setup. If other memory types are required
...
+
+   * - data_size
+     - 0x4
+     - 0x4
+     - The size (in bytes) of the CDI
+
+   * - profile_id
+     - 0x4
+     - 0x8
+     - the Open DICE profile identifier (See :numref:`opendice_profile_id` for the allowed values)
```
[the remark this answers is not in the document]
thomas-fossati: If I `make html` the reference is generated correctly.

For the avoidance of doubt, I will change this into a reference to the table itself.

source/transfer_list.rst
thomas-fossati: @danh-arm @manish-pandey-arm @jmarinho @apalos

I believe I have addressed all the outstanding comments except for this one.

manish-pandey-arm: I would prefer it to be in generic range starting at 0x7 as Mbed-TLS heap has now been moved to TF-A range (0x105) https://github.com/FirmwareHandoff/firmware_handoff/pull/45/files

jmarinho:
Agreed with @thomas-fossati to close this PR for now as we do not have imediate use-cases needing it.
We will re-open the PR later when that changes.

PR #42, opened 2024-03-25
Fix tag ID of RW memory layout entry

manish-pandey-arm:
Tag ID of RW memory layout entry was wrongly put as "104" instead of "0x104".


sjg20 (approved):
Perhaps we should change the spec to say that all values are in hex?

danh-arm (approved):

PR #45, opened 2024-06-13
Add Mbed-TLS crypto heap info entry

harrisonmutai-arm:
This PR introduces a TE to communicate details about the memory allocated for Mbed-TLS’s dynamic memory management. The buffer, described by the TE, is zero-initialised and intended for internal stack-based memory allocation by the library [1]. It's address and size are used by subsequent layers during the intialisation of Mbed-TLS.

[1] https://arm-software.github.io/CMSIS-mbedTLS/latest/memory__buffer__alloc_8h.html#ac70d134be54133c272d8eab2cb85dfbf

### DIFF
--- a/source/transfer_list.rst
+++ b/source/transfer_list.rst
@@ -820,7 +820,6 @@ and reserved for use as a TPM Command Response Buffer interface.
      - Size of CRB.
 
 
-
 .. _tf_entries:
 
 Entries related to Trusted Firmware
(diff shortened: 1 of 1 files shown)


source/transfer_list.rst
```diff
@@ -819,6 +819,46 @@ and reserved for use as a TPM Command Response Buffer interface.
      - hdr_size + 0x8
      - Size of CRB.
 
+Cryptogragphy library heap information (XFERLIST_CRYPTO_LIB_HEAP_INFO)
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+The cryptograph library heap info specifies the base address and size of a
+region of memory on the heap that has been carved out for use by a cryptography
+library.
+
```
apalos, line 827: I don't think the description is enough. Some information needs to be added on how the subsequent layers are supposed to use that information. e.g do they have to reserve that memory when they define their heap etc.

sjg20: +1

Also, to me, 'heap' suggests some sort of memory allocation going on. Is this in a particular format? How would someone set it up? Is this targeted at a particular library? If so you could reference it here.

sjg20 (changes requested):

source/transfer_list.rst
harrisonmutai-arm: Thank you both for your valuable feedback!

Our primary focus is on the Mbed-TLS library. At this point, I haven’t found any other libraries where this information might be pertinent. Given this, I’ve refined the scope of the TE to exclusively cover Mbed-TLS and relocated the TE within TF-A’s range. Please share your thoughts if you think there’s potential for the TE in the generic range.

Additionally, I’ve enhanced the explanation regarding the expected usage of memory by the producer and consumer. This should provide a clearer understanding of how this TE should be used.

manish-pandey-arm (approved):

manish-pandey-arm:
@sjg20 @apalos Are you happy with the wording?

sjg20 (approved):

PR #46, opened 2024-09-13
Improve HTML Documentation Testing and Deployment

harrisonmutai-arm:
Ensure HTML documentation is consistently built and published from the latest commit on the main branch or on pull requests. Implement automated workflows to streamline the testing and deployment process.


.github/workflows/main.yml
```diff
@@ -1,27 +1,54 @@
 name: Build specification
 on:
   push:
-    branches: [ main ]
```
jmarinho, line 4: nit: just want to make sure this is intended.

jmarinho (approved):
LGTM, but need others to ACK

.github/workflows/main.yml
harrisonmutai-arm: My linter seems to have auto-formatted this and a few other places. It has no real effect, but I am happy to split the bits that were refactored out into a seperate patch if you prefer?

manish-pandey-arm (approved):

PR #49, opened 2024-10-22
Allow N entries of same type, if sub-type defined

jmarinho:
Add TL rules:
- TE types may allow multiple entries, provided that they contain a sub-type field.
- For all the TE types that do not allow multiple entries, the TL must contain at most a TE of that type.

Change-Id: Ie88ce9f1eb7efb9065bd6f9059739e363f84055d

### DIFF
--- a/source/transfer_list.rst
+++ b/source/transfer_list.rst
@@ -42,7 +42,6 @@ Transfer list requirements
 
 **R4:** When relocating the TL, the offset from `tl_base_pa` to the nearest alignment boundary specified by the `alignment` field in the TL header must be preserved.
 
-
 Transfer list header
 --------------------
 
@@ -193,8 +192,7 @@ they're always expected to appear together on the first platform adding the tag
 (to encourage reusability in different situations). Alternatively, complex data
 may be represented in a different kind of well-established handoff data
 structure (e.g. FDT [DT]_, HOB [PI]_) that is inserted into the TL as a single
-TE. The same tag ID may occur multiple times in the TL to represent multiple
-instances of the same kind of object. Tag layouts (including the meaning of all
+TE. Tag layouts (including the meaning of all
 fields) are considered stable after being added to this specification and must
 never be changed in a backwards-incompatible way. If a backwards-incompatible
 change is desired, a new tag ID should be allocated for the new version of the
@@ -210,6 +208,13 @@ TEs should not contain explicit version numbers and instead just use the
 `data_size` value to infer how many fields exist. TE layouts which have been
 changed like this must clearly document which fields or valid values were added
 at a later time, and in what order.
+
+It is strongly recommended that entries of a given type are
+unique in the TL. If firmware designs require multiple TE instances of a given
+type, then that TE type definition should provide sufficient information for
+the TE consumer to disambiguate between all TE instances. That information can
+be, for example, a sub-type field or contained within a self-describing data
+blob.
 
 The TL must not hold pointers or addresses within its entries, which refer to
 anything in the TL. These can make it difficult to relocate the TL. TL


manish-pandey-arm:
Should we provide a guideline on how the sub-type field should be defined ? Also, does it make sense to add sub-type ID to XFERLIST_EXEC_EP_INFO64 which can helpful in setting the example. IMHO its worth to have a paragraph explaining how to add sub-type ID and how the implementation can handle it

jwerner-chromium:
Hi Jose, can you add some reasoning to this change (or link if that discussion took place elsewhere)? I don't really see the point of specifying that there has to be a type subfield in this spec when the location and format of that subfield isn't specified. What difference does it make to the toplevel spec? (E.g. in terms of API design it still means that you need a way to walk all TEs of the same tag and pass off interpretation to the tag-specific code.)

I'm also not sure I understand the point of type subfields to begin with. If there are different kinds of content that you want to store, wouldn't you just make them separate tags? The tag space is intentionally so large that people should not shy away from just allocating as much as they need.  Did someone have a practical need for so many tags that you didn't want them to allocate them all in the main tag space?

jmarinho:
> Hi Jose, can you add some reasoning to this change (or link if that discussion took place elsewhere)? I don't really see the point of specifying that there has to be a type subfield in this spec when the location and format of that subfield isn't specified. What difference does it make to the toplevel spec? (E.g. in terms of API design it still means that you need a way to walk all TEs of the same tag and pass off interpretation to the tag-specific code.)
> 
> I'm also not sure I understand the point of type subfields to begin with. If there are different kinds of content that you want to store, wouldn't you just make them separate tags? The tag space is intentionally so large that people should not shy away from just allocating as much as they need. Did someone have a practical need for so many tags that you didn't want them to allocate them all in the main tag space?

Hi @jwerner-chromium,

This topic was discussed a few times in the past, between varying groups of people.
From my observation of those discussion, it's unanimous that TEs, of a given type, should be unique in a TL.

There are some use-cases where uniqueness is not practical, though.
For instance, last year we defined  the XFERLIST_EXEC_EP_INFO64 TE, to represent the entry point of a given FW image.
There can be multiple FW images and thus multiple entry points.

This class of TEs should be an exception.
The PR spells that out, and also tries to provide minimal guidance for the design/definition of TE types that allow multiple instances.

jmarinho:
> Should we provide a guideline on how the sub-type field should be defined ? Also, does it make sense to add sub-type ID to XFERLIST_EXEC_EP_INFO64 which can helpful in setting the example. IMHO its worth to have a paragraph explaining how to add sub-type ID and how the implementation can handle it

Hi @manish-pandey-arm 

On the first question, I don't think the spec should be prescriptive on how the sub-type ID field is defined.
Different TEs may want to have their own pattern.
The consequence is code traversing the TL must return a TE of a given type at a time.
We cannot implement a get_function taking "tag_id" and "sub-type" as arguments.

On the second question: (I think) It makes a lot of sense to add the sub0type field in XFERLIST_EXEC_EP_INFO64.

Let's see what others say on this topic.

source/transfer_list.rst
```diff
@@ -42,6 +42,9 @@ Transfer list requirements
 
 **R4:** When relocating the TL, the offset from `tl_base_pa` to the nearest alignment boundary specified by the `alignment` field in the TL header must be preserved.
 
+**R5:** If a TE type definition contains a sub-type field, then multiple entries of that TE type are allowed in the TL. The TE must have unique values in the sub-type field.
```
sjg20, line 45: perhaps *a* sub-type, rather than *the* ?

This seems reasonable to me, but I can think of a case where a subtype per se might not be needed.

If each entry has a memory range, then the range itself (e.g. start and size) would be enough to distinguish the entries. But still, I think the language here is an improvement, so I am happy enough with it.

sjg20 (approved):

jwerner-chromium:
> For instance, last year we defined the XFERLIST_EXEC_EP_INFO64 TE, to represent the entry point of a given FW image. There can be multiple FW images and thus multiple entry points.

Well, we _could_ have different TEs for each image type, right? Or is that a legacy structure that's already in use somewhere else and already contains a subfield type, and that's why you want to reuse it that way?

In general, I'm just not sure I see the benefit of mandating this if the subfield itself is not "visible" to the generic spec anyway. If you said "each TE may only appear once no matter what", then that comes with clear API design benefits. But if you say "TEs may only appear once except for these cases", I think you might as well leave it completely up to the TE-implementer how often they may appear under what circumstances, there's no real added benefit from proscribing certain variations.

Also, if I have that TE that e.g. describes a memory region that's reserved for some reason, and it contains an offset and a size value, can I have multiple of that because the offset is technically a sort of "subtype field" (every address being a different "type" of memory reservation)? Or does that not count? And if it doesn't, what's the rationale in saying that if someone wants to define entry points for different firmware images they may use the same tag for that, but if they want to define different reserved memory regions they may not? Is it really useful to draw that distinction? And does it make sense that my hypothetical reserved memory TE would suddenly become "legal" if I just added a third "reservation type" field to it, even if that field wasn't actually needed by any of its consumers and is just there to make the spec happy?

I kinda feel like an all-or-nothing decision would make more sense here. Either no TEs may be duplicate so that APIs can actually offer a find_tag(tag_id) API without having to deal with the multiple return headaches, or you might as well leave it up to the TE implementers to come up with whatever works best for their use case.

Also, from just the text arrangement side, I think the "Transfer List Requirements" is just there to specify some core principles about the generic TL structure itself, while all the rules and guidelines about how to design TEs are described in the "TL contents" section further down, so I think this requirement would fit better in there. (Note that it currently says "The same tag ID may occur multiple times in the TL to represent multiple instances of the same kind of object", so if you are changing that principle you should update the text there to match.)

manish-pandey-arm:
To answer @jwerner-chromium earlier question 
> Did someone have a practical need for so many tags that you didn't want them to allocate them all in the main tag space?
One of the usecases we discussed in the past was related with XFERLIST_FDT mentioned by @apalos where a platform might have multiple device trees which need to be passed by earlier boot phases and the one to be used will be decided at later stage. Having a sub-type will help in choosing one out of many device trees.

sjg20:
> To answer @jwerner-chromium earlier question
> 
> > Did someone have a practical need for so many tags that you didn't want them to allocate them all in the main tag space?
> 
> One of the usecases we discussed in the past was related with XFERLIST_FDT mentioned by @apalos where a platform might have multiple device trees which need to be passed by earlier boot phases and the one to be used will be decided at later stage. Having a sub-type will help in choosing one out of many device trees.

Well I'm actually opposed to that and I believe that this is the wrong approach. There should just be one devicetree.

Perhaps this would be better wording:

**R5:** If a TE type definition contains a sub-type field, or some other means of distinguishing one TE from another using their contents, then multiple entries of that TE type are allowed in the TL.

i.e. I don't believe it needs to be a sub-type. For a memory range, it could just be that the ranges are different:

TE1:  <100 200>
TE2:  <400 600>

danh-arm:
I'm coming to this late but FWIW I'm against this change. I think there are reasonable use-cases for multiple TEs of the same type, even without defining a sub-type field, especially for TEs that are self-describing.

Note the spec currently states:
"The same tag ID may occur multiple times in the TL to represent multiple instances of the same kind of object", so this would also  need changing if we changed the policy.

This is perhaps one for our next meeting to close out.

danh-arm (changes requested):
I disapprove generally of this change

jmarinho:
> I disapprove generally of this change

Rephrased this PR as guidance.
We want to allow multiple instances of the same type (when justified). In those cases TE types should contain enough info to distinguish between different TE instances.

source/transfer_list.rst
```diff
@@ -42,6 +42,8 @@ Transfer list requirements
 
 **R4:** When relocating the TL, the offset from `tl_base_pa` to the nearest alignment boundary specified by the `alignment` field in the TL header must be preserved.
 
+It is strongly recommended that entries of a given type are unique in the TL.
```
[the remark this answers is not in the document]
jmarinho: Moved the recommendation to Section 2.4.

source/transfer_list.rst
```diff
@@ -193,8 +192,12 @@ they're always expected to appear together on the first platform adding the tag
 (to encourage reusability in different situations). Alternatively, complex data
 may be represented in a different kind of well-established handoff data
 structure (e.g. FDT [DT]_, HOB [PI]_) that is inserted into the TL as a single
-TE. The same tag ID may occur multiple times in the TL to represent multiple
-instances of the same kind of object. Tag layouts (including the meaning of all
+TE. It is strongly recommended that entries of a given type are unique in the
```
danh-arm, line 195: This strong recommendation should only be the case if the TE definition does not meet the requirements you list below. Just move to the end of the paragraph?

danh-arm (changes requested):

sjg20 (approved):

PR #50, opened 2024-10-22
Clarify that TransferList has its roots in BlobList

jmarinho:
Change-Id: I12c0a72a9920239f093d6eeef871c4d8c25cd4b5
Signed-off-by: Jose Marinho <jose.marinho@arm.com>


manish-pandey-arm (approved):

sjg20 (approved):
The only addition is that U-Boot uses the bloblist for internal tables, not just for passing information

Issue #51, opened 2024-11-19
Standard data type for firmware binaries


J-Alves:
Hi,

In the context of the Hafnium project, we are exploring the use of Transfer List structures to package the artefacts of each partition. Hafnium is the reference implementation for S-EL2 firmware for an aarch64 architecture, following the FF-A standard.
The idea is to load a TL per partition agregating the artefacts the SPMC needs to process. Each TL will be signed with independent keys from each provider to enforce the trusted boot flow; hence, there is a need for a TL package per SP.

Currently, the same effect is achieved with an impdef package, which is not very scalable. TLs seem appealing due to the well-curated specification, availability of a firmware library and build tool in TF-A which can be used.

The data types required are:
- FDT.
- Partitions Binary.
- HOB List (optional).

The FDT and HOB types are part of the standard data types.
I couldn't find a standard data type that could be used for the partition binary. Presently, using the custom data type range.

I would like to obtain feedback from the community on the idea of adding the binary data type. 
Any objections as to why there shouldn't be such a type?
If there are no objections, what would be the best way to do the respective update to the specification?

Best regards,
João Alves

mathias-arm:
Different use case but similar needs. What do you think about #9 and #10?

J-Alves:
Hi,

Thanks for providing context to those.

IIUC, the GUID would discriminate between the different SP binaries.

This is not needed in my case, as we have the FF-A UUID as the identifier for the SP at the system level. It also seems to imply a sanity check on the GUID within the SPMC. I don't think It seems unnecessary.

Let me know if I am missing something.

Cheers

jwerner-chromium:
The general idea is that if none of the existing tags are suitable for your use case, you should just send a PR to add a new tag in the standardized ID range with whatever format you need. You can start a new block for tags related to Hafnium (e.g. at 0x200). See https://github.com/FirmwareHandoff/firmware_handoff/blob/main/source/transfer_list.rst#entry-type-allocation for details.  If the format proves useful enough other projects may choose to reuse it later, and if not that's fine too.

You should not be using tags in the custom (non-standardized) range (starting 0xfff000) for anything you expect to use long-term in open-source code.

danh-arm:
[...]

PR #52, opened 2024-12-07
FF-A manifest and SP binary TE types

J-Alves:
The full rationale can be found in issue [51](https://github.com/FirmwareHandoff/firmware_handoff/issues/51).

This pull request concerns adding a dedicated data type for an SP binary and making the SPMC manifest TE type more generic, to be used as the SP's FF-A manifest.

This allows the Hafnium SPMC to use the TL format to package SP's artefacts.

### DIFF
--- a/source/transfer_list.rst
+++ b/source/transfer_list.rst
@@ -1032,6 +1032,44 @@ into its memory map during platform setup. If other memory types are required
      - hdr_size + 0x8
      - The size of the memory region.
 
+**DT formatted FF-A manifest entry layout (XFERLIST_DT_FFA_MANIFEST)**
+
+This entry type holds the FF-A manifest image whice is in DT format [DT]_,
+as described in [TFAFFAMB]_.
+This manifest contains the SP (Secure Partition) configuration, consumed
+by the SPMC at boot time.
+
+It may also contain some information to the SP itself.
+
+.. _tab_dt_ffa_manifest:
+.. list-table:: DT formatted FF-A manifest type layout
+   :widths: 2 2 2 8
+
+   * - Field
+     - Size (bytes)
+     - Offset (bytes)
+     - Description
+
+   * - tag_id
+     - 0x3
+     - 0x0
+     - The tag_id field must be set to **0x106**.
+
+   * - hdr_size
+     - 0x1
+     - 0x3
+     - |hdr_size_desc|
+
+   * - data_size
+     - 0x4
+     - 0x4
+     - The size of FF-A manifest in bytes.
+
+   * - ffa_manifest
+     - data_size
+     - hdr_size
+     - Holds a FF-A manifest image in DT format.
+
 Mbed-TLS heap information (XFERLIST_MBEDTLS_HEAP_INFO)
 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
 
@@ -1072,8 +1110,6 @@ passed to later stages for intialisation of Mbed-TLS.
      - 0x8
      - hdr_size + 0x8
      - Size of memory region.
-
-
 
 .. |hdr_size_desc| replace:: The size of this entry header in bytes must be set to **8**.
 .. |current_version| replace:: `0x1`
--- a/source/transfer_list.rst
+++ b/source/transfer_list.rst
@@ -951,6 +951,43 @@ software running in Secure, Non-Secure, or Realm modes.
      - hdr_size
      - Holds a single `entry_point_info` structure.
 
+**FF-A SP binary (XFERLIST_FFA_SP_BINARY)**
+
+This entry holds a reference to an FF-A Secure Partition (SP) binary.
+
+This TE type is for an SPMC implementation to identify which entry
+relates to the SP's binary, such that it can install the binary and
+hand-over execution.
+
+.. _tab_ffa_sp_binary:
+.. list-table:: An FF-A SP binary type layout
+   :widths: 2 2 2 8
+
+   * - Field
+     - Size (bytes)
+     - Offset (bytes)
+     - Description
+
+   * - tag_id
+     - 0x3
+     - 0x0
+     - The tag_id field must be set to **0x103**.
+
+   * - hdr_size
+     - 0x1
+     - 0x3
+     - |hdr_size_desc|
+
+   * - data_size
+     - 0x4
+     - 0x4
+     - The size of the SP binary in bytes.
+
+   * - ffa_sp_binary
+     - data_size
+     - hdr_size
+     - Holds the FF-A SP binary.
+
 **Read-Write Memory Layout Entry Layout (XFERLIST_RW_MEM_LAYOUT64)**
 
 This entry type holds a structure that describes the layout of a read-write
--- a/source/transfer_list.rst
+++ b/source/transfer_list.rst
@@ -826,7 +826,7 @@ Entries related to Trusted Firmware
 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
 
 The following entry types are defined for Trusted Firmware projects,
-including TF-A and OP-TEE:
+including TF-A, OP-TEE and Hafnium:
 
 **OP-TEE pageable part address entry layout (XFERLIST_OPTEE_PAGEABLE_PART_ADDR)**
 
@@ -868,9 +868,11 @@ the OP-TEE OS.
 **DT formatted SPMC manifest entry layout (XFERLIST_DT_SPMC_MANIFEST)**
 
 This entry type holds the SPMC (Secure Partition Manager Core) manifest image
-which is in DT format [DT]_ and described in [TFAFFAMB]_.
+which is in DT format [DT]_ and described in [SPMCATTR]_.
 This manifest contains the SPMC attribute node consumed by the SPMD
 (Secure Partition Manager Dispatcher) at boot time.
+It may also contain some information for the SPMC implementation, to
+initialize itself.
 
 .. _tab_dt_spmc_manifest:
 .. list-table:: DT formatted SPMC manifest type layout
--- a/source/references.rst
+++ b/source/references.rst
@@ -28,3 +28,5 @@ References
 .. [TCG_EFI] `Trusted Computing Group EFI Protocol Specification <https://trustedcomputinggroup.org/resource/tcg-efi-protocol-specification>`__
 
 .. [TF_BL31] `Data structures used in the BL31 cold boot interface <https://trustedfirmware-a.readthedocs.io/en/latest/design/firmware-design.html#data-structures-used-in-the-bl31-cold-boot-interface>`__
+
+.. [SPMCATTR] `The SPMC manifest: <https://hafnium.readthedocs.io/en/latest/secure-partition-manager/secure-partition-manager.html#spmc-manifest>`__


source/transfer_list.rst
```diff
@@ -993,6 +995,48 @@ into its memory map during platform setup. If other memory types are required
...
+
+This entry holds a reference to an FF-A secure partition's binary.
+
+Hafnium SPMC supports the SP's artefacts to be packaged as a Transfer List
+generated at build time (with TF-A's tlc tool).
+
+The respective Transfer List contents: SP binary/image, the FF-A manifest, and
+other resoures (e.g. HOB - optional).
+
+This entry type helps identify the executable that SPMC should bootstrap/hand-over
```
harrisonmutai-arm, line 1008: Could you provide more context or clarify what you're trying to say? It would be helpful to explain how this entry type is intended to be used, for the benefit of others who may be unsure.

harrisonmutai-arm (changes requested):

source/transfer_list.rst
```diff
@@ -993,6 +995,48 @@ into its memory map during platform setup. If other memory types are required
      - hdr_size + 0x8
      - The size of the memory region.
 
+**FF-A SP binary (XFERLIST_FFA_SP_BINARY)**
+
+This entry holds a reference to an FF-A secure partition's binary.
```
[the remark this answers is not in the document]
J-Alves: Done

source/transfer_list.rst
```diff
@@ -993,6 +995,48 @@ into its memory map during platform setup. If other memory types are required
...
+
+   * - Field
+     - Size (bytes)
+     - Offset (bytes)
+     - Description
+
+   * - tag_id
+     - 0x3
+     - 0x0
+     - The tag_id field must be set to **0x105**.
```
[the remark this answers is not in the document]
J-Alves: Done

source/transfer_list.rst
```diff
@@ -993,6 +995,48 @@ into its memory map during platform setup. If other memory types are required
...
+generated at build time (with TF-A's tlc tool).
+
+The respective Transfer List contents: SP binary/image, the FF-A manifest, and
+other resoures (e.g. HOB - optional).
+
+This entry type helps identify the executable that SPMC should bootstrap/hand-over
+execution.
+
+.. _tab_ffa_sp_binary:
+.. list-table:: DT formatted SPMC manifest type layout
```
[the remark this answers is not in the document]
J-Alves: Done

source/transfer_list.rst
```diff
@@ -951,6 +951,48 @@ software running in Secure, Non-Secure, or Realm modes.
      - hdr_size
      - Holds a single `entry_point_info` structure.
 
+**FF-A SP binary (XFERLIST_FFA_SP_BINARY)**
+
+This entry holds a reference to an FF-A secure partition's binary.
+
+Hafnium SPMC supports the SP's artefacts to be packaged as a Transfer List
+generated at build time (with TF-A's tlc tool). The respective Transfer List
+contents: SP binary/image, the FF-A manifest, and other optionl resoures
```
odeprez, line 960: typos => optionl, resoures

source/transfer_list.rst
```diff
@@ -951,6 +951,48 @@ software running in Secure, Non-Secure, or Realm modes.
...
+(e.g. HOB).
+
+Under the the described context, this TE type is for an SPMC implementation
```
odeprez, line 963: fix  'the the'

source/transfer_list.rst
```diff
@@ -894,12 +896,12 @@ This manifest contains the SPMC attribute node consumed by the SPMD
...
      - 0x4
      - 0x4
-     - The size of SPMC manifest in bytes.
+     - The size of the DTB in bytes.
 
    * - spmc_man
      - data_size
      - hdr_size
-     - Holds a SPMC manifest image in DT format.
+     - Holds a SPMC or FF-A partition manifest image in DT format.
```
danh-arm, line 904: Holds an FF-A manifest image in DT format.

source/transfer_list.rst
```diff
@@ -951,6 +951,48 @@ software running in Secure, Non-Secure, or Realm modes.
      - hdr_size
      - Holds a single `entry_point_info` structure.
 
+**FF-A SP binary (XFERLIST_FFA_SP_BINARY)**
+
+This entry holds a reference to an FF-A secure partition's binary.
```
danh-arm, line 956: When secure partition is first used, you should capitalize and add (SP), then use SP from then on. Also SP needs spelling out in the commit message.

source/transfer_list.rst
```diff
@@ -951,6 +951,48 @@ software running in Secure, Non-Secure, or Realm modes.
...
+
+Hafnium SPMC supports the SP's artefacts to be packaged as a Transfer List
+generated at build time (with TF-A's tlc tool). The respective Transfer List
+contents: SP binary/image, the FF-A manifest, and other optionl resoures
+(e.g. HOB).
```
danh-arm, line 961: I think this whole paragraph should be in the commit message, not the spec itself. Also, you could say that there are already TEs defined for FF-A manifest and HOB list, but not for an SP binary.

source/transfer_list.rst
danh-arm: Suggest removing "Under the the described context," (if you accept my other comment).

source/transfer_list.rst
```diff
@@ -951,6 +951,48 @@ software running in Secure, Non-Secure, or Realm modes.
...
+
+Under the the described context, this TE type is for an SPMC implementation
+to identify which entry relates to the SP's binary, such that it can
+bootstrap/hand-over execution.
```
danh-arm, line 965: perhaps could be a bit more descriptive, e.g. "so that it can install the binary and hand-over execution to it".

danh-arm (changes requested):

source/transfer_list.rst
```diff
@@ -865,12 +865,14 @@ the OP-TEE OS.
...
      - Holds the address of OP-TEE pageable part
 
-**DT formatted SPMC manifest entry layout (XFERLIST_DT_SPMC_MANIFEST)**
+**DT formatted FF-A manifest entry layout (XFERLIST_DT_FFA_MANIFEST)**
 
-This entry type holds the SPMC (Secure Partition Manager Core) manifest image
-which is in DT format [DT]_ and described in [TFAFFAMB]_.
-This manifest contains the SPMC attribute node consumed by the SPMD
+This entry type holds an FF-A manifest, which can either be an SPMC (Secure Partition Manager Core)
+manifest image, or an FF-A partition's manifest. Both shall be in DT format [DT]_.
```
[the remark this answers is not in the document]
J-Alves: At the moment there are only those two kinds of manifest.
They are different, hence the rewrite.

Only the FF-A SP manifest relates to the contents documented in the TFAFFAMB reference.
SPMC manifest: https://hafnium.readthedocs.io/en/latest/secure-partition-manager/secure-partition-manager.html#spmc-manifest
I will add a reference to the later one as well.

source/transfer_list.rst
```diff
@@ -865,12 +865,14 @@ the OP-TEE OS.
...
+In case it relates to an SPMC manifest, it contains the SPMC attribute node consumed by the SPMD
 (Secure Partition Manager Dispatcher) at boot time.
+In case it relates to a FF-A partition manifest, it contains partition's information following
+the FF-A manifest binding specification [TFAFFAMB]_.
```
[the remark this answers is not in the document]
J-Alves: See my previous comment.
The SPMC does not follow that specification.

I was trying to correct the statement.

source/transfer_list.rst
J-Alves: Done

source/transfer_list.rst
```diff
@@ -894,12 +896,12 @@ This manifest contains the SPMC attribute node consumed by the SPMD
    * - data_size
      - 0x4
      - 0x4
-     - The size of SPMC manifest in bytes.
+     - The size of the DTB in bytes.
```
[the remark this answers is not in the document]
J-Alves: Done

source/transfer_list.rst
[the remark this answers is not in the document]
danh-arm: OK, so the existing spec points to the wrong reference. Given that the 2 manifests use 2 different specs, I've changed my mind about making the existing TE more generic. I now think the existing SPMC TE should have its reference fixed and then add a new one for the SP manifest. Sorry, for messing you around but I wasn't aware of this difference before.

source/transfer_list.rst
```diff
@@ -865,12 +865,15 @@ the OP-TEE OS.
...
-This entry type holds the SPMC (Secure Partition Manager Core) manifest image
-which is in DT format [DT]_ and described in [TFAFFAMB]_.
-This manifest contains the SPMC attribute node consumed by the SPMD
-(Secure Partition Manager Dispatcher) at boot time.
+This entry type holds an FF-A manifest, which can either be an SPMC (Secure Partition Manager Core)
+manifest image, or an FF-A Secure Partition (SP) manifest image.
+The SPMC manifest contains the SPMC *attribute* node consumed by the SPMD (Secure Partition Manager
+Dispatcher) at boot time [SPMCATTR]_.
+The FF-A SP manifest has the partition's information following the FF-A manifest binding specification
+[TFAFFAMB]_.
```
danh-arm, line 875: Given that the 2 manifests use 2 different specs, I've changed my mind about making the existing TE more generic. I now think the existing SPMC TE should have its reference fixed and then add a new one for the SP manifest.

danh-arm: It's not so much about the risk of these manifests existing in the same TL, but that they represent different logical objects. It's true that a side-effect of using the DT format is that it enables easy detection of the content, but we've had strong feedback previously that we should not overload TEs that use DT format with different content. Otherwise, we could just use the existing XFERLIST_FDT for all these uses of DT formatted objects. Or even replace this spec with DT!

Actually, the preference is to not overload use of DT at all and use TEs defined in this spec if possible. That might be too disruptive in this case but hopefully gives you an idea of the direction of travel.

J-Alves:
@danh-arm the current change created another entry for the FF-A manifest, and corrected the reference to the SPMC manifest documentation.
I hope this fits better with your previous comments.

source/transfer_list.rst
```diff
@@ -868,9 +868,10 @@ the OP-TEE OS.
 **DT formatted SPMC manifest entry layout (XFERLIST_DT_SPMC_MANIFEST)**
 
 This entry type holds the SPMC (Secure Partition Manager Core) manifest image
-which is in DT format [DT]_ and described in [TFAFFAMB]_.
+which is in DT format [DT]_ and described in [SPMCATTR]_.
 This manifest contains the SPMC attribute node consumed by the SPMD
 (Secure Partition Manager Dispatcher) at boot time.
+It may also contain some information to the SPMC implementation itself.
```
danh-arm, line 874: nit: s/to/about/

danh-arm (approved):
nit: Could add "Fixes #51" to the last commit message.
nit: "For the moment, Hafnium as the S-EL2 SPMC is used". Not sure this comment is relevant in this spec and may age badly.

Only minor comments overall so I approve.

source/transfer_list.rst
```diff
@@ -1034,7 +1072,43 @@ passed to later stages for intialisation of Mbed-TLS.
...
+     - 0x1
+     - 0x3
+     - |hdr_size_desc|
+
+   * - data_size
+     - 0x4
+     - 0x4
+     - The size of FF-A manifest in bytes.
+
+   * - spmc_man
```
danh-arm, line 1108: Just spotted that this name needs changing, e.g. ffa_manifest

source/transfer_list.rst
```diff
@@ -1034,7 +1072,43 @@ passed to later stages for intialisation of Mbed-TLS.
      - hdr_size + 0x8
      - Size of memory region.
 
+**DT formatted SPMC manifest entry layout (XFERLIST_DT_FFA_MANIFEST)**
```
danh-arm, line 1075: I've just noticed the "Mbed-TLS heap information" TE incorrectly starts a new section. This issue was already there but means this TE appears in the wrong section. Would be good if you can submit a fix for that too.

danh-arm (changes requested):

source/transfer_list.rst
J-Alves: Corrected the sentence.

source/transfer_list.rst
J-Alves: I will move it back.
The intention was to retain order of TE ID.
It seems HEAP is 0x105

J-Alves:
Thanks again @danh-arm !

harrisonmutai-arm (approved):

manish-pandey-arm (approved):

manish-pandey-arm (approved):

manish-pandey-arm (approved):

source/transfer_list.rst
```diff
@@ -951,6 +951,43 @@ software running in Secure, Non-Secure, or Realm modes.
...
+
+   * - data_size
+     - 0x4
+     - 0x4
+     - The size of the SP binary in bytes.
+
+   * - ffa_sp_binary
+     - data_size
+     - hdr_size
+     - Holds the FF-A SP binary.
```
sjg20, line 989: What is the format of this binary?

source/transfer_list.rst
```diff
@@ -1032,6 +1032,44 @@ into its memory map during platform setup. If other memory types are required
...
+
+   * - data_size
+     - 0x4
+     - 0x4
+     - The size of FF-A manifest in bytes.
+
+   * - ffa_manifest
+     - data_size
+     - hdr_size
+     - Holds a FF-A manifest image in DT format.
```
sjg20, line 1071: DT might be better as devicetree? But so long as there is DT elsewhere, this is fine.

Is this format per the link here?:

https://hafnium.readthedocs.io/en/latest/secure-partition-manager/secure-partition-manager.html#spmc-manifest

If so, can you please mention it here, with a reference, or similar.

Also, that link seems to imply that the DT is partially the same as normal DT schema, but with some additions. I see an 'attribute' node with no compatible string and property names which use underscore when they should use hyphen. Also, some mention should be made there of the address-cells property being 2?

You might also be interested in the Universal Payload project:

https://universalpayload.github.io/spec/index.html

sjg20 (changes requested):

danh-arm (approved):

source/transfer_list.rst
danh-arm: Hi @sjg20. The DT acronym is defined earlier, although more the doc generally could be more consistent on its use of acronyms (that doesn't need to be fixed here)
This TE follows the format in https://trustedfirmware-a.readthedocs.io/en/latest/components/ffa-manifest-binding.html, which is already referenced.
Your feedback on those reference docs is useful, thanks.
Finally we should discuss the UP project sometime but that's beyond the scope of this PR.

So I think we're good to go?

sjg20 (approved):

PR #57, opened 2025-02-06
Relax language around use of non-standardized tag id range

danh-arm:
Slightly relax the language around the use of the non-standardized tag id range, to allow the use between closely coupled components that must only be used within a particular firmware implementation. In some cases, the TE definition is continuously evolving or very complex to define, which makes standardization an onerous requirement.

### DIFF
--- a/source/transfer_list.rst
+++ b/source/transfer_list.rst
@@ -246,13 +246,18 @@ reservations of tag ranges for specific use.
 The {0xff_f000, ..., 0xff_ffff} range is reserved for non-standardized use.
 Anyone is free to use tags from that range for any custom TE layout without
 adding their definitions to this specification first. The use of this range is
-*strongly discouraged* for anything other than local experiments or code that
-will only ever be used in closed-source components owned by the entity
-controlling the entire final firmware image. In particular, a creator of a TE
-that just contains platform-specific data or internal structures specific to
-a single firmware implementation, should allocate a standardized tag for it
-in this specification -- using the non-standardized range is strongly
-discouraged. Since standards often emerge organically, the goal is to
+*strongly discouraged* for anything other than:
+
+* local experiments,
+
+* code that will only ever be used in closed-source components owned by the
+  entity controlling the entire final firmware image, or
+
+* TEs internal to a single component or transferred between closely coupled
+  components in a given firmware implementation.
+
+Even in the latter case, if the TE definition is stable the non-standardized
+range is discouraged. Since standards often emerge organically, the goal is to
 create unique tag IDs for everything just in case it turns out to be useful in
 more applications than initially anticipated. Basically, whenever you're
 submitting code for a new TE layout to any public open-source project, that's


danh-arm:
Hi @jwerner-chromium. You might want to check this one as you wrote the original words. The issue is that we're struggling to define the TE we use for the platform specific config shared by TF-A BL1 and BL2. It contains a lot of evolving policy and chain of trust information, that will never be of any use outside these components.

sjg20:
I think it is better to make the effort to document things.

This seems to defeat the purpose of this spec. If we start seeing secret things which must be there for the system to boot, but cannot be generated by open source code, then how can we maintain interoperability?

Perhaps a better understanding of the current challenge you have would help?

jmarinho:
> I think it is better to make the effort to document things.
> 
> This seems to defeat the purpose of this spec. If we start seeing secret things which must be there for the system to boot, but cannot be generated by open source code, then how can we maintain interoperability?
> 
> Perhaps a better understanding of the current challenge you have would help?

@manish-pandey-arm , @danh-arm can you provide more details on the use-case you have in mind?
Is this related to a datastructure that's available upstream and  documented in the TF-A docs (with guarantees of stability)?

danh-arm:
Hi @sjg20. I agree it's better to document things where possible and I think I retained the essence of the guidance to this effect. But I think there are legitimate uses of the non-standardized range. The use-case we're struggling with relates to PR #37 (now closed). The information in this Trusted Boot Firmware Config is complex, platform specific, unstable and only used by TF-A BL1/BL2. It's not even passed between BL1 and BL2, but loaded/authenticated by both, with only a small subset of the data used by BL1. The data certainly won't make it's way to subsequent firmware stages. As mentioned in that PR, where possible we'll create and document TEs for some of this data in case it's more widely usable, but some of the data (e.g. the Chain of Trust) is likely to evolve more rapidly than this spec can cope with, especially when each TF-A platform may have its own distinct CoT. I think it's reasonable for this to remain non-standardized indefinitely.

jwerner-chromium:
I would say that even in that case you should probably allocate a tag in the standardized range for it. If you're loading something in both BL1 and BL2, you'll probably want to version the struct somehow so that if anyone ever mixes and matches different BL1 and BL2 builds together for some reason, they can identify a mismatch in the struct if there is one. The tag allocation rules just serve to remind you to do that versioning (using the tag ID as a struct version) and I think that's a good thing. We have 16 million tag IDs so I don't think we need to worry about burning a few dozen because this struct is frequently evolving. Or if you are worried that that would become a problem, it is still perfectly legal to allocate a single tag in the standardized range, then have the first field in that structure be an internal struct version, and make all the rest of the fields (and the length of the entire struct) dependent on that (in a way that can be expanded in the future by updating the spec and using previously reserved internal struct versions). The only thing that's not allowed in the standardized range is serializing data without any clear way to identify how the data inside should be interpreted at all, which I think is a bad idea in general and a good thing for the spec to discourage.

danh-arm:
> I would say that even in that case you should probably allocate a tag in the standardized range for it.
We tried that in PR #37 but it stalled due to difficulties in defining the TE in the spec. We're happy to resurrect that if it's the consensus.

> Or if you are worried that that [number of tags] would become a problem...
No, we're not worried about that. Just the churn in the actual definition. The definition would link back to TF-A documentation. Note, we expect to continue using DT format indefinitely for at least some part of this as it's easier to manage changes using that than defining flat structs. Although we expect these TF-A internal DT bindings to change, we think it will be rare for this to cause incompatibility between BL1 and BL2. If that does happen, I agree using a new tag ID would be a good idea.

If we can agree that's OK to link back to TF-A documentation and let TF-A determine when compatibility is broken, then we'll go back to the previous approach and I'll withdraw this PR.

jwerner-chromium:
FWIW I think it's perfectly fine to have the contents of a TE defined by a link to a description elsewhere, especially if they are already standardized elsewhere and more structured than raw C struct fields (e.g. an FDT like in this case). We already have examples of that in the spec, e.g. with the ACPI table tag that just links to the ACPI spec.

I think the important part is that the linked information must remain at that address permanently and must not change in a backwards-incompatible way (same rules as for TE contents defined inline). For this specific case, it looks like you're including a top-level `compatible` prop in that FDT which identifies it. So as long as you only add optional new fields and keep the meaning of the existing ones the same you should be fine, and if you want to change or remove existing fields you could update that compatible prop value to signify that change (maybe make it a little more explicit in the doc that implementations must check that value before interpreting the other fields). Then I think this should count as a backwards-compatible stable interface fit for inclusion as a TE.

source/transfer_list.rst
```diff
@@ -246,13 +246,18 @@ reservations of tag ranges for specific use.
...
-that just contains platform-specific data or internal structures specific to
-a single firmware implementation, should allocate a standardized tag for it
-in this specification -- using the non-standardized range is strongly
-discouraged. Since standards often emerge organically, the goal is to
+*strongly discouraged* for anything other than:
+
+* local experiments,
+
+* code that will only ever be used in closed-source components owned by the
+  entity controlling the entire final firmware image, or
```
sjg20, line 254: IMO that defeats the purpose of this specification, which is to allow interoperable components.

There are only two cases, really:
1. The undocumented information is necessary for booting, iwc the device is a brick without those secret pieces. There is no point in putting it in this spec and it should certainly not be allocated a tag
2. The undocumented information is not necessary for booting, iwc we should leave it out.

sjg20 (changes requested):

source/transfer_list.rst
sjg20: > HI @sjg20. As stated [here](https://github.com/FirmwareHandoff/firmware_handoff/pull/57#issuecomment-2657193037), the latest proposal is to withdraw this PR and resurrect #37 with improved TF-A documentation of the bindings (including a stable link). Julius said [here](https://github.com/jwerner-chromium) he was OK with this approach. Are you? It's really not an option for us to turn this into a giant C struct and document here.

If it is a giant C struct then yes it doesn't belong in Firmware Handoff, which it for simple things. It should use the devicetree.

sjg20:
I hope it is OK to close this for now, as I apparently got confused about it still being an open PR.

PR #61, opened 2025-03-12
Fix ep-Info field offsets and add 64-bit fields

harrisonmutai-arm:
Both `XFERLIST_DT_FFA_MANIFEST` and `XFERLIST_EXEC_EP_INFO32` were erroneously assigned the tag ID 0x106. Since `XFERLIST_EXEC_EP_INFO32` is newer and not yet widely used, its tag ID is updated to 0x107 to resolve the conflict.

Additionally, fix the field layout for the 32-bit EP_INFO entry: register offsets were misaligned due to an incorrect header size assumption. Each member is adjusted by 0x4 to align correctly with the actual size of param_header_h.

Also, add the layout definition for the 64-bit EP_INFO entry type to complete the specification.

### DIFF
--- a/source/transfer_list.rst
+++ b/source/transfer_list.rst
@@ -1176,37 +1176,38 @@ subsequent images. It's usage is identical to the 64-bit form represented by
    * - ep_info_hdr
      - 0x8
      - hdr_size
-     - Header of type :ref:`param_header<tab_param_header>` containing structure
-       version, size, and attributes.
+     - Header of type :ref:`param_header<tab_param_header>` containing the
+       information about this structure. The type must be `0x1`, version `0x2`,
+       and size `0x24`.
 
    * - pc
      - 0x4
-     - hdr_size + 0x4
+     - hdr_size + 0x8
      - Program counter (entry point into image).
 
    * - spsr
      - 0x4
-     - hdr_size + 0x8
+     - hdr_size + 0xc
      - Saved Program Status Register.
 
    * - lr_svc
      - 0x4
-     - hdr_size + 0xc
+     - hdr_size + 0x10
      - Link register.
 
    * - r0
      - 0x4
-     - hdr_size + 0x10
+     - hdr_size + 0x14
      - Register R0.
 
    * - r1
      - 0x4
-     - hdr_size + 0x14
+     - hdr_size + 0x18
      - Register R1.
 
    * - r2
      - 0x4
-     - hdr_size + 0x18
+     - hdr_size + 0x1c
      - Register R2.
 
 The structures header contains an attributes field which is used to encode the image's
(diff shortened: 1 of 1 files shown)


source/transfer_list.rst
```diff
@@ -973,6 +973,93 @@ software running in Secure, Non-Secure, or Realm modes.
      - hdr_size
      - Holds a single `entry_point_info` structure.
```
danh-arm, line 974: Should be removed now you're explicitly inlining the ep_info fields.

source/transfer_list.rst
```diff
@@ -973,6 +973,93 @@ software running in Secure, Non-Secure, or Realm modes.
...
 
+   * - ep_info_hdr
+     - 0x8
+     - hdr_size
+     - Header of type :ref:`param_header<tab_param_header>` containing structure
+       version, size, and attributes.
```
danh-arm, line 980: The actual version used should be quoted here. I'm not sure if the attributes should be initialized in any particular way for this structure type but would be good to check. I notice the param_header size field and the data_size field are effectively duplicated info (with a fixed offset from each other). The actual size used should be quoted for one of them (I think stating the data_size would be clearer).

Ditto for for the 32-bit version. 

In future we could potentially support new versions of the ep_info if it were done in a BC way and all changes were documented.

danh-arm (changes requested):
The "Add 64-bit ep-info fields" commit could do with some content explaining that you're inlining the specific ep_info fields for clarity.

source/transfer_list.rst
harrisonmutai-arm: > The actual version used should be quoted here.

If I’ve understood correctly, these two entries should refer to a specific version throughout their lifetime. So we should explicitly state the supported version here, along with its associated structure sizes. Is that right? Also, how do you see us maintaining backward compatibility in practice?

> I'm not sure if the attributes should be initialized in any particular way for this structure type but would be good to check.

Good point. I think it’s worth noting that the attributes field should be initialized to match the security state of the entry point—otherwise, it defaults to RES0.

danh-arm: > So we should explicitly state the supported version here, along with its associated structure sizes. Is that right?

Yes, that's right.

> Also, how do you see us maintaining backward compatibility in practice?

If we make a compatible change in future (e.g. adding a field to version Y), we could say "The version field must be either X or Y", "The size must be A for version X or B for version Y", and the new field description could say "only supported in version Y".

danh-arm (approved):

jmarinho:
> Sorry for the last minute change @danh-arm, @jmarinho pointed out there's a collision in the 32-bit ep info tag id. Since we are cleaning up the layout decided we may as well resolve conflict.

Minor note: the tag-id clash was reported by Levi (Yeoreum Yun).

danh-arm (approved):

PR #65, opened 2025-03-17
Add nightly job to test doc builds and links

### LINKED ISSUE #64: Extend CI to detect broken links
Would be good to detect these automatically.

harrisonmutai-arm:
Create a scheduled job to run nightly, ensuring documentation builds and links are validated and errors are caught early.

Fixes #64

### DIFF
--- a/.github/workflows/daily.yml
+++ b/.github/workflows/daily.yml
@@ -1,0 +1,30 @@
+name: Nightly Sphinx Build and Link Check
+
+on:
+  schedule:
+    - cron: "0 2 * * *"
+  # This allows the job to be manually triggered.
+  workflow_dispatch:
+
+jobs:
+  build-and-check-links:
+    name: "Build PDF and HTML documentation"
+    runs-on: ubuntu-24.04
+    steps:
+      - uses: actions/checkout@v4
+      - uses: actions/setup-python@v5
+      - name: Install dependencies
+        run: |
+          sudo apt-get update
+          sudo apt-get install -y librsvg2-bin \
+            python3-sphinx latexmk texlive-latex-extra \
+            python3-sphinxcontrib.svg2pdfconverter sphinx-multiversion \
+            python3-sphinx-rtd-theme
+
+      - name: Build Sphinx documentation
+        run: |
+          make html latexpdf
+
+      - name: Check for broken links
+        run: |
+          sphinx-build -b linkcheck source/ _build/linkcheck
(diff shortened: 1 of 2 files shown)


danh-arm:
Looks good, but what will happen when this build fails? Is there a notification to maintainers or similar?

harrisonmutai-arm:
At the moment, only the person who creates the job receives a notification. Unfortunately, GitHub doesn’t provide a built-in mechanism for broadcasting workflow failures [1]—this appears to be a long-standing issue with no resolution in sight. The only workaround I’ve found is that users can manually subscribe to receive notifications when a job fails [2].

One alternative could be adding a nightly badge—similar to the one in the README—to make the nightly status visible to anyone visiting the repository.

[1] https://github.com/orgs/community/discussions/18039
[2] https://docs.github.com/en/actions/monitoring-and-troubleshooting-workflows/monitoring-workflows/notifications-for-workflow-runs

danh-arm:
> At the moment, only the person who creates the job receives a notification.

Hmm, that's not ideal.

> The only workaround I’ve found is that users can manually subscribe to receive notifications when a job fails.

It's not good enough to only be able to subscribe to a single run.

> One alternative could be adding a nightly badge—similar to the one in the README—to make the nightly status visible to anyone visiting the repository.

If that's not too difficult then great!

Anyway, my comments don't block merging of this PR - it's still better than nothing.

danh-arm (dismissed):

harrisonmutai-arm:
> It's not good enough to only be able to subscribe to a single run.

I meant you can subscribe to receive notifications for the worklow. That is, you'd receive notifications for any failures. That's at least what I understood the documentation to mean.

> If that's not too difficult then great!

Not at all, added a follow up commit with a badge for the daily. I'm hoping this does the job but it's hard to say since the worklow will only show up once this PR is merged. 

image

danh-arm:
> I meant you can subscribe to receive notifications for the worklow. That is, you'd receive notifications for any failures. That's at least what I understood the documentation to mean.

I read it differently but now I'm not sure! It says "If you enable email or web notifications for GitHub Actions, you'll receive a notification when any workflow runs _that you've triggered_ have completed", which implies to me that people won't be notified of jobs they haven't triggered. It then says "Notifications for scheduled workflows are sent to the user who initially created the workflow", which implies to me that others won't be notified about scheduled jobs either.

> [I] added a follow up commit with a badge for the daily. I'm hoping this does the job but it's hard to say since the workflow will only show up once this PR is merged.

Cool. That works for me.

danh-arm (approved):

harrisonmutai-arm:
> "Notifications for scheduled workflows are sent to the user who initially created the workflow", which implies to me that others won't be notified about scheduled jobs either.

Ah you're probably right :( 

Their documentation isn’t very clear. This page [1] suggests you can set up notifications for repositories you’re watching with Actions enabled, but it’s not obvious whether the same restriction applies—that is, whether you only get notifications for Actions you’ve triggered yourself or created. 

[1] https://docs.github.com/en/account-and-profile/managing-subscriptions-and-notifications-on-github/setting-up-notifications/configuring-notifications#github-actions-notification-options

jmarinho (approved):
LGTM

sjg20 (approved):

Issue #62, opened 2025-03-17
Add instructions for "Removing a TE" and "Overwriting a TE"


danh-arm:
In the "Standard operations" section, it would be good to add instructions for "Removing a TE" and "Overwriting a TE". These are not necessarily trivial operations and currently the only reference to this is in the XFERLIST_VOID section. The instructions should state that implementations may need to scrub any padding bytes left from the old TE if these contain secret data that receivers should not see.

manish-pandey-arm:
PR #66 takes care of this issue

PR #66, opened 2025-03-20
Update standard operations section

manish-pandey-arm:
Provide details around security considerations and add new standard operations like "Removing a TE" and "Overwriting a TE" as mentioned in https://github.com/FirmwareHandoff/firmware_handoff/issues/62

### DIFF
--- a/source/transfer_list.rst
+++ b/source/transfer_list.rst
@@ -392,23 +392,13 @@ Inputs:
    starting at `te_base_addr` to `tl.checksum`.
 
 #. If an existing XFERLIST_VOID TE was chosen to be overwritten in step 1, and
-   `old_void_data_size - new_data_size` is greater or equal to `0x8`:
+   `old_void_data_size - new_data_size` is greater than or equal to `0x8`, then
+   create a new void TE to fill the remaining space by calling `Adding a void TE`_
+   with the following arguments:
 
-   #. Use `te_base_addr + align8(new_data_size + 0x8)` as the new `te_base_addr`
-      for a new XFERLIST_VOID tag.
+   #. `void_te.base_addr` = `te_base_addr + align8(new_data_size + 0x8)`
 
-   #. If `has_checksum`, subtract the sum of the 8 bytes from `te_base_addr` to
-      `te_base_addr + 0x8` from `tl.checksum`.
-
-   #. Set `te.tag_id` (`te_base_addr + 0x0`) to `0x0` (XFERLIST_VOID).
-
-   #. Set `te.hdr_size` (`te_base_addr + 0x3`) to `0x8`.
-
-   #. Set `te.data_size` (`te_base_addr + 0x4`) to
-      `old_void_data_size - align8(new_data_size) - 0x8`.
-
-   #. If `has_checksum`, add the sum of the 8 bytes from `te_base_addr` to
-      `te_base_addr + 0x8` to `tl.checksum`.
+   #. `void_te.data_size` =  `old_void_data_size - align8(new_data_size + 0x8)`
 
 Adding a new TE with special data alignment requirement
 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
@@ -510,6 +500,32 @@ Inputs:
 .. note::
    After relocating a TL, implementations should consider scrubbing the old TL memory if it contains
    any secrets that might be accessible to later untrusted software.
+
+Helper Routines
+^^^^^^^^^^^^^^^
+
+Adding a void TE
+~~~~~~~~~~~~~~~~
+
+Inputs:
+
+- `te_base_addr`: Base address where void TE to be added
+- `data_size`: Size in bytes of the data to be encapsulated in void TE
+
+#. If `has_checksum`, subtract the sum of `te.hdr_size + data_size` bytes starting at
+   `te_base_addr` from `tl.checksum`.
+
+#. Set `te.tag_id` (`te_base_addr + 0x0`) to `0x0` (XFERLIST_VOID)
+
+#. Set `te.hdr_size` (`te_base_addr + 0x3`) to `0x8`
+
+#. Set `te.data_size` (`te_base_addr + 0x4`) to `align8(data_size)`
+
+#. *(optional)* Set the `data_size` bytes starting at `te_base_addr + te.hdr_size` to 0x0
+
+#. If `has_checksum`, add the sum of `te.hdr_size + data_size` bytes starting at
+   `te_base_addr` with `tl.checksum`
+
 
 .. _sec_std_entries:
 
--- a/source/transfer_list.rst
+++ b/source/transfer_list.rst
@@ -292,6 +292,11 @@ This section describes the valid operations that may be performed on a TL in
 more detail, in order to clarify how to use the various fields and to serve as a
 guideline for implementation.
 
+.. note::
+This section assumes te.hdr_size is 0x8 for all TE entries. This is used consistently throughout the examples.
+In the future, if entries with different header sizes are introduced, some operations may need to explicitly
+take te.hdr_size as an argument.
+
 Validating a TL header
 ^^^^^^^^^^^^^^^^^^^^^^
 
@@ -338,8 +343,7 @@ Inputs:
       is smaller or equal to `tl.used_size`, otherwise abort (the TL is corrupted).
 
    #. If `te.tag_id` (`te_base_addr + 0x0`) is a known tag, interpret the data
-      at `te_base_addr + te.hdr_size` accordingly. (Do not hardcode the value
-      for `te.hdr_size`, even for known tags!) Otherwise, ignore the tag and
+      at `te_base_addr + te.hdr_size` accordingly. Otherwise, ignore the tag and
       proceed with the next step.
 
    #. Add `align8(te.hdr_size + te.data_size)` to `te_base_addr`.
@@ -362,14 +366,14 @@ Inputs:
 
    #. Use the `te_base_addr` of this tag for the rest of the operation.
 
-   #. If `has_checksum`, Subtract the sum of `align8(new_data_size + 0x8)` bytes
+   #. If `has_checksum`, Subtract the sum of `align8(new_data_size + te.hdr_size)` bytes
       starting at `te_base_addr` from `tl.checksum`.
 
    #. Skip the next step (step 2) with all its substeps.
 
 #. Calculate `te_base_addr` as `tl_base_addr + tl.used_size`.
 
-   #. If `tl.total_size - tl.used_size` is smaller than `align8(new_data_size + 0x8)`,
+   #. If `tl.total_size - tl.used_size` is smaller than `align8(new_data_size + te.hdr_size)`,
       abort (not enough room to add TE).
 
    #. If `has_checksum`, subtract the sum of the 4 bytes from
@@ -386,19 +390,19 @@ Inputs:
 
 #. Set `te.data_size` (`te_base_addr + 0x4`) to `new_data_size`.
 
-#. Copy or generate the TE data into `te_base_addr + 0x8`.
+#. Copy or generate the TE data into `te_base_addr + te.hdr_size`.
 
-#. If `has_checksum`, add the sum of `align8(new_data_size + 0x8)` bytes
+#. If `has_checksum`, add the sum of `align8(new_data_size + te.hdr_size)` bytes
    starting at `te_base_addr` to `tl.checksum`.
 
 #. If an existing XFERLIST_VOID TE was chosen to be overwritten in step 1, and
-   `old_void_data_size - new_data_size` is greater than or equal to `0x8`, then
+   `old_void_data_size - new_data_size` is greater than or equal to `te.hdr_size`, then
    create a new void TE to fill the remaining space by calling `Adding a void TE`_
    with the following arguments:
 
-   #. `void_te.base_addr` = `te_base_addr + align8(new_data_size + 0x8)`
+   #. `void_te.base_addr` = `te_base_addr + align8(new_data_size + te.hdr_size)`
 
-   #. `void_te.data_size` =  `old_void_data_size - align8(new_data_size + 0x8)`
+   #. `void_te.data_size` =  `old_void_data_size - align8(new_data_size + te.hdr_size)`
 
 Removing a TE
 ^^^^^^^^^^^^^
@@ -411,7 +415,7 @@ Inputs:
 
    #. `void_te.base_addr` = `te_base_addr`
 
-   #. `void_te.size` = `te.data_size + te.hdr_size - 0x8`
+   #. `void_te.size` = `te.data_size`
 
 #. *(Optional)* Implementations may perform memory management by inspecting adjacent entries
    and coalescing consecutive `XFERLIST_VOID` entries into a single larger one. This can help
@@ -443,12 +447,12 @@ Inputs:
 #. If `has_checksum`, add the sum of `align8(new_data_size)` bytes starting at
    `te_base_addr + te.hdr_size` to `tl.checksum`.
 
-#. If `te.data_size - align8(new_data_size)` is greater than or equal to `0x8`, create a new
+#. If `te.data_size - align8(new_data_size)` is greater than or equal to `te.hdr_size`, create a new
    void TE to fill the remaining space by calling `Adding a void TE`_ with the following arguments:
 
    #. `te_base_addr` = `te_base_addr + align8(new_data_size + te.hdr_size)`
 
-   #. `data_size` = `te.data_size - align8(new_data_size - 0x8)`
+   #. `data_size` = `te.data_size - align8(new_data_size - te.hdr_size)`
 
 Adding a new TE with special data alignment requirement
 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
--- a/source/transfer_list.rst
+++ b/source/transfer_list.rst
@@ -507,6 +507,9 @@ Inputs:
 #. If `has_checksum`, add the sum of the 4 bytes from `new_tl_base + 0xc` to
    `new_tl_base + 0x10` to `tl.checksum` (`new_tl_base + 0x4`).
 
+.. note::
+   After relocating a TL, implementations should consider scrubbing the old TL memory if it contains
+   any secrets that might be accessible to later untrusted software.
 
 .. _sec_std_entries:
 
--- a/source/transfer_list.rst
+++ b/source/transfer_list.rst
@@ -417,6 +417,39 @@ Inputs:
    and coalescing consecutive `XFERLIST_VOID` entries into a single larger one. This can help
    reduce fragmentation and improve reuse of space in the Transfer List.
 
+Overwriting a TE
+^^^^^^^^^^^^^^^^
+
+Inputs:
+
+- `te_base_addr`: Base address of the Transfer Entry (TE) to be overwritten
+- `new_data_size`: Size in bytes of the new data to be encapsulated in the TE
+- [data]: Data to be copied into the TE
+
+#. If `te.data_size` is smaller than `new_data_size`, the overwrite operation must return an error.
+   The caller is then responsible for:
+
+   #. Reclaiming the existing space by calling `Adding a void TE`_ with `te_base_addr` and `te.data_size`.
+
+   #. Adding a new TE by invoking `Adding a new TE`_ with appropriate arguments.
+
+#. If `has_checksum`, subtract the sum of `te.data_size` bytes starting at
+   `te_base_addr + te.hdr_size` from `tl.checksum`.
+
+#. Set `te.data_size` (`te_base_addr + 0x4`) to `align8(new_data_size)`.
+
+#. Copy or generate the new TE data into `te_base_addr + te.hdr_size`.
+
+#. If `has_checksum`, add the sum of `align8(new_data_size)` bytes starting at
+   `te_base_addr + te.hdr_size` to `tl.checksum`.
+
+#. If `te.data_size - align8(new_data_size)` is greater than or equal to `0x8`, create a new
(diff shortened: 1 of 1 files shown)


source/transfer_list.rst
```diff
@@ -391,23 +408,60 @@ Inputs:
...
+- `tag_id`: ID number of the tag to be deleted
+- `scrub_data`: Boolean option to scrub TE data
+
+#. Validate the TL header as mentioned in `Validating a TL header`_
+
+#. Locate the `te_base_addr` of the TE to be removed by iterating through the TL.
+
+#. Mark the TE as void by setting its tag_id to XFERLIST_VOID
+
+#. Erase the TE data if `scrub_data` is TRUE
```
jwerner-chromium, line 433:
```
*(optional)* Set the `te.data_size` bytes starting at `te_base_addr + te.hdr_size` to 0x0.
```

source/transfer_list.rst
```diff
@@ -391,23 +408,60 @@ Inputs:
...
+Inputs:
+
+- `tl_base_addr`: Base address of the TL from which TE to be deleted
+- `tag_id`: ID number of the tag to be deleted
+- `scrub_data`: Boolean option to scrub TE data
+
+#. Validate the TL header as mentioned in `Validating a TL header`_
+
+#. Locate the `te_base_addr` of the TE to be removed by iterating through the TL.
+
```
jwerner-chromium, line 430: To keep the checksum instructions in line with other operations, add here (before making changes):
```
If `has_checksum`, xor the `te.hdr_size + te.data_size` bytes starting at `te_base_addr` with `tl.checksum`
```

source/transfer_list.rst
```diff
@@ -391,23 +408,60 @@ Inputs:
...
+#. Mark the TE as void by setting its tag_id to XFERLIST_VOID
+
+#. Erase the TE data if `scrub_data` is TRUE
+
+#. If checksum is enabled, recalculate and update `tl.checksum`
```
jwerner-chromium, line 435:
```
If `has_checksum`, xor the `te.hdr_size + te.data_size` bytes starting at `te_base_addr` with `tl.checksum`
```

source/transfer_list.rst
```diff
@@ -391,23 +408,60 @@ Inputs:
    `te_base_addr` with `tl.checksum`.
 
 #. If an existing XFERLIST_VOID TE was chosen to be overwritten in step 1, and
-   `old_void_data_size - new_data_size` is greater or equal to `0x8`:
+   `old_void_data_size - new_data_size` is greater or equal to `0x8` then call
+   `Adding a void TE`_ with following arguments:
+
+   #. `void.te.base_addr` = `te_base_addr + align8(new_data_size + 0x8)`
```
jwerner-chromium, line 414: nit: `void_te`, not `void.te`

source/transfer_list.rst
```diff
@@ -391,23 +408,60 @@ Inputs:
...
+
+   #. `void.te.size` =  `old_void_data_size - align8(new_data_size)`
```
jwerner-chromium, line 416: This is missing another `- 0x8`

source/transfer_list.rst
```diff
@@ -343,6 +343,23 @@ Inputs:
 
    #. Add `align8(te.hdr_size + te.data_size)` to `te_base_addr`.
 
+
+Adding a void TE
```
jwerner-chromium, line 347: I would call this "Overwriting part of the TL with a void TE", because this isn't really "adding" a TE (to the end) the way the normal "Adding a new TE" operation does.

I would maybe put this at the end or at least further down in this section, because it's not really a useful operation on its own, it's only a common step for two others, and it's weird for the reader to see this first.

source/transfer_list.rst
```diff
@@ -343,6 +343,23 @@ Inputs:
...
+^^^^^^^^^^^^^^^^
+
+Inputs:
+
+- `te_base_addr`: Base address where void TE to be added
```
jwerner-chromium, line 352: It might be useful to put the alignment part in here so that it doesn't need to be mentioned in the two calling operations (e.g. call this `start_address`, and then start with calculating `te_base_addr` as `align8(start_address)`)?

source/transfer_list.rst
```diff
@@ -343,6 +343,23 @@ Inputs:
...
+- `size`: Total size available for void TE
+
```
jwerner-chromium, line 354: You also need to xor the checksum here, to first remove the old state of those bytes from the checksum before adding the new one.

source/transfer_list.rst
```diff
@@ -343,6 +343,23 @@ Inputs:
...
+#. Set `te.tag_id` (`te_base_addr + 0x0`) to `0x0` (XFERLIST_VOID)
+
+#. Set `te.hdr_size` (`te_base_addr + 0x3`) to `0x8`
+
+#. Set `te.data_size` (`te_base_addr + 0x4`) to `align8(size) - 0x8`
+
```
jwerner-chromium, line 360: If you want to mention optional data erasure in the "Remove" operation, you should probably do the same here?  (Note that this will change the area over which the checksum needs to be calculated. In fact, it may be easier to rewrite the "Remove" operation by referencing this one?)

source/transfer_list.rst
```diff
@@ -391,23 +408,60 @@ Inputs:
...
+Inputs:
+
+- `tl_base_addr`: Base address of the TL from which TE to be overwritten
+- `tag_id`: ID number of the tag to be overwritten
+- `new_data_size`: Size in bytes of the data to be encapsulated in the TE
+- [data]: Data to be copied into the TE
+
+#. Validate the TL header as mentioned in :ref:`Validating a TL header`
+
+#. Locate the `te_base_addr` of the TE to be overwritten by iterating through the TL
```
jwerner-chromium, line 449: Same as above I think this should be the input, and it's odd to mention the validation step here.

source/transfer_list.rst
```diff
@@ -391,23 +408,60 @@ Inputs:
...
+
+#. Ensure that `te.data_size` is greater or equal to `new_data_size`
```
jwerner-chromium, line 451: Rather than just write this I think we should give guidance on how to handle the case when it isn't. I think you could write:
```
#. If `te.data_size` is smaller than `new_data_size`, abort this operation here and instead first follow `Adding a void TE` with `te_base_addr` and `te.data_size`, then follow `Adding a new TE` with `tl_base_addr`, `tag_id`, `new_data_size` and `[data]`.
```

source/transfer_list.rst
```diff
@@ -391,23 +408,60 @@ Inputs:
...
 
-   #. Use `te_base_addr + align8(new_data_size + 0x8)` as the new `te_base_addr`
-      for a new XFERLIST_VOID tag.
+#. Erase the old TE data from `te_base_addr + 0x8` until  `te.data_size`
```
jwerner-chromium, line 453: This is inefficient. If erasure is (optionally) desired, we only need to overwrite the rest of the old data that remains after new data has been written. I think this should be intergrated into the "Add a void TE" operation, because it is the same requirement in all its callers.

source/transfer_list.rst
```diff
@@ -391,23 +408,60 @@ Inputs:
...
 
-   #. If `has_checksum`, xor the 8 bytes from `te_base_addr` to
-      `te_base_addr + 0x8` with `tl.checksum`.
+#. Copy the TE data into `te_base_addr + 0x8`
```
jwerner-chromium, line 455: nit: "Copy or generate" to match the wording in "Adding a new TE"

source/transfer_list.rst
jwerner-chromium, line 451: After this step you need to first subtract the old header and data bytes from the checksum if `has_checksum`.

source/transfer_list.rst
```diff
@@ -391,23 +408,60 @@ Inputs:
...
-   #. Set `te.hdr_size` (`te_base_addr + 0x3`) to `0x8`.
+   #. `void.te.base_addr` = `te_base_addr + align8(new_data_size + 0x8)`
 
-   #. Set `te.data_size` (`te_base_addr + 0x4`) to
-      `old_void_data_size - align8(new_data_size) - 0x8`.
+   #. `void.te.size` =  `te.data_size - align8(new_data_size)`
 
-   #. If `has_checksum`, xor the 8 bytes from `te_base_addr` to
-      `te_base_addr + 0x8` with `tl.checksum`.
+#. If checksum is enabled, recalculate and update `tl.checksum`
```
jwerner-chromium, line 464: This should again explicitly describe how to account for the new bytes after making the change.

jwerner-chromium (dismissed):

source/transfer_list.rst
danh-arm: I kind of agree with this but would go one step further. If we're adding an "Overwriting a TE" section then I think this it's unnecessary to have this operation also embedded within the existing "Adding a TE" section. So I would remove that bit from "Adding a TE", rename it to "Append a TE", and inline this "Overwriting part of the TL with a void TE" operation within the "Overwriting a TE" section.

danh-arm (changes requested):

source/transfer_list.rst
danh-arm: On 2nd thoughts, you might still need this operation in 2 places ("Overwriting a TE" and "Removing a TE"), in which case it might better in its own section. But still it should be below the operations that use it.

source/transfer_list.rst
```diff
@@ -391,23 +408,60 @@ Inputs:
...
+
+   #. `void.te.size` =  `old_void_data_size - align8(new_data_size)`
+
+Removing a TE
+^^^^^^^^^^^^^
+
+Inputs:
+
+- `tl_base_addr`: Base address of the TL from which TE to be deleted
+- `tag_id`: ID number of the tag to be deleted
```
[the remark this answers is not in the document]
manish-pandey-arm: It is implicitly used in finding "te_base_addr", rephrased the sentence.

source/transfer_list.rst
```diff
@@ -391,23 +408,60 @@ Inputs:
...
+- `scrub_data`: Boolean option to scrub TE data
+
+#. Validate the TL header as mentioned in `Validating a TL header`_
```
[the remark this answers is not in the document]
manish-pandey-arm: Done

source/transfer_list.rst
```diff
@@ -391,23 +408,60 @@ Inputs:
...
+
+#. Locate the `te_base_addr` of the TE to be removed by iterating through the TL.
```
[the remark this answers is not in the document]
manish-pandey-arm: Done

source/transfer_list.rst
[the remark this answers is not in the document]
manish-pandey-arm: used te_base_addr as input

source/transfer_list.rst
```diff
@@ -391,23 +391,52 @@ Inputs:
...
+- `tl_base_addr`: Base address of the TL from which TE to be overwritten
+- `te_base_addr`: Base address of the TE to be overwritten
+- `new_data_size`: Size in bytes of the data to be encapsulated in the TE
+- [data]: Data to be copied into the TE
+
+#. If `te.data_size` is smaller than `new_data_size`, abort this operation here and instead first
+   follow `Adding a void TE`_ with `te_base_addr` and `te.data_size`, then follow `Adding a new TE`_
+   with `tl_base_addr`, `tag_id`, `new_data_size` and `[data]`
+
+#. If `has_checksum`, xor the `te.data_size` bytes starting at `te_base_addr + te.hdr_size` with `tl.checksum`
```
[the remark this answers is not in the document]
manish-pandey-arm: My understanding is, An overwrite operation updates the data linked to a tag ID without changing the tag ID itself.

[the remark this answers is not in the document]
manish-pandey-arm: Thanks for spotting, done!

source/transfer_list.rst
```diff
@@ -391,23 +391,52 @@ Inputs:
...
+
+#. Copy `new_data_size` byte of data starting at `te_base_addr + te.hdr_size`
 
-   #. Set `te.hdr_size` (`te_base_addr + 0x3`) to `0x8`.
+#. If `has_checksum`, xor the `new_data_size` bytes starting at `te_base_addr + te.hdr_size` with `tl.checksum`
```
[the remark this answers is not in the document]
manish-pandey-arm: The API descriptions currently use 0x8 for existing TEs (e.g., VOID_TE), based on their current header size. However, since this may not hold true for all TEs in the future, the generic descriptions refer to te.hdr_size to avoid assumptions. If preferred, I can submit a separate PR to replace all uses with the fixed 0x8 to ensure consistency.

source/transfer_list.rst
```diff
@@ -398,6 +398,19 @@ Inputs:
 
    #. `void_te.size` =  `old_void_data_size - align8(new_data_size + 0x8)`
 
+Removing a TE
```
sjg20, line 401: Seems to be missing a commit message

sjg20 (dismissed):

source/transfer_list.rst
manish-pandey-arm: Have only got commit header but no body, hope that is fine.

jmarinho (dismissed):

source/transfer_list.rst
sjg20: Please add a body to all commits.

source/transfer_list.rst
```diff
@@ -400,6 +400,19 @@ Inputs:
...
+
+Inputs:
+
+- `te_base_addr`: Base address of the TE to be removed
+
+#. Invoke `Adding a void TE`_ with following arguments
+
+   #. `void_te.base_addr` = `te_base_addr`
+
+   #. `void_te.size` = `te.data_size + te.hdr_size - 0x8`
```
jmarinho: is the second arg void_te.data_size ?

see https://github.com/FirmwareHandoff/firmware_handoff/pull/66/commits/5f526c05259352757d7d7a8c1b4fc7db97e40820#diff-75d1d8518f52351181570a42b6cee154a72b2eb15b3771b969f182b2e537b41eR513

source/transfer_list.rst
```diff
@@ -413,6 +413,35 @@ Inputs:
...
+#. Set `te.data_size` (`te_base_addr + 0x4`) to `align8(new_data_size)`
+
+#. Copy or generate the new TE data into `te_base_addr + te.hdr_size`
+
+#. If `has_checksum`, add the `new_data_size` bytes starting at `te_base_addr + te.hdr_size` with `tl.checksum`
+
+#. If `te.data_size - new_data_size` is greater than or equal to `0x8`, then create a new
+   void TE to fill the remaining space by calling `Adding a void TE`_ with following arguments:
+
+   #. `void.te.base_addr` = `te_base_addr + te.hdr_size + align8(new_data_size)`
```
jmarinho: nit: void_te ?

source/transfer_list.rst
```diff
@@ -413,6 +413,35 @@ Inputs:
 
    #. `void_te.size` = `te.data_size + te.hdr_size - 0x8`
 
+Overwriting a TE
+^^^^^^^^^^^^^^^^
+
+Inputs:
+
+- `tl_base_addr`: Base address of the TL from which TE to be overwritten
```
jmarinho: nit: I'd argue tl_base_addr is not an argument to the Overwrite TE operation.
The tl_base_addr is used in the "Adding a new TE" procedure, which is executed if the "Overwriting a TE" aborts.

jmarinho (dismissed):
small set of nits, but otherwise LGTM

source/transfer_list.rst
```diff
@@ -507,6 +539,33 @@ Inputs:
...
+
+Adding a void TE
+~~~~~~~~~~~~~~~~
+
+Inputs:
+
+- `te_base_addr`: Base address where void TE to be added
+- `data_size`: Size in bytes of the data to be encapsulated in void TE
+
+#. If `has_checksum`, subtract `te.hdr_size + data_size` bytes starting at `te_base_addr` from `tl.checksum`
```
harrisonmutai-arm: Should this say "subtract the sum of the bytes starting at `te_base_addr` up to `te_base_addr + te.hdr_size + data_size` from `tl.checksum` instead? Otherwise, this reads as if were subtracting the sum of the header size and data size.

source/transfer_list.rst
```diff
@@ -507,6 +539,33 @@ Inputs:
...
+
+#. Set `te.tag_id` (`te_base_addr + 0x0`) to `0x0` (XFERLIST_VOID)
```
harrisonmutai-arm, line 572: I'm in two minds about providing the offsets explicitly (e.g. `te_base_addr + 0x3`). On one hand it's a useful reference for the implementer, but it may prove a bit of a burden to maintain down the line. As I understand it, the header layout is liable to change. If it does, we'd have to fix all the places in the spec where we reference hard-coded offsets. Given the structure is well defined, I think it's probably sufficient to just refer to it directly (i.e. `te.tag_id`).

source/transfer_list.rst
```diff
@@ -392,23 +392,55 @@ Inputs:
...
-      `te_base_addr + 0x8` from `tl.checksum`.
+   #. `void_te.size` =  `old_void_data_size - align8(new_data_size + 0x8)`
 
-   #. Set `te.tag_id` (`te_base_addr + 0x0`) to `0x0` (XFERLIST_VOID).
+Removing a TE
+^^^^^^^^^^^^^
+
+Inputs:
+
+- `te_base_addr`: Base address of the TE to be removed
```
harrisonmutai-arm, line 412: For libtl we use the `tl_base_addr` to perform very naive memory management. If the previous and subsequent TE's are empty we try to coalesce them with the removed TE. This is arguably an optional implementation detail though.

source/transfer_list.rst
```diff
@@ -392,23 +392,55 @@ Inputs:
...
+- `tl_base_addr`: Base address of the TL from which TE to be overwritten
+- `te_base_addr`: Base address of the TE to be overwritten
+- `new_data_size`: Size in bytes of the data to be encapsulated in the TE
+- [data]: Data to be copied into the TE
+
+#. If `te.data_size` is smaller than `new_data_size`, abort this operation here and instead first
+   follow `Adding a void TE`_ with `te_base_addr` and `te.data_size`, then follow `Adding a new TE`_
+   with `tl_base_addr`, `tag_id`, `new_data_size` and `[data]`
+
+#. If `has_checksum`, subtract the `te.data_size` bytes starting at `te_base_addr + te.hdr_size` from `tl.checksum`
```
harrisonmutai-arm: Similar comments to below, should this say "sum of `te.data_size` bytes.

source/transfer_list.rst
```diff
@@ -392,23 +392,55 @@ Inputs:
...
 
 #. If an existing XFERLIST_VOID TE was chosen to be overwritten in step 1, and
-   `old_void_data_size - new_data_size` is greater or equal to `0x8`:
+   `old_void_data_size - new_data_size` is greater than or equal to `0x8`, then
+   create a new void TE to fill the remaining space by calling `Adding a void TE`_
+   with the following arguments:
 
-   #. Use `te_base_addr + align8(new_data_size + 0x8)` as the new `te_base_addr`
-      for a new XFERLIST_VOID tag.
+   #. `void_te.base_addr` = `te_base_addr + align8(new_data_size + 0x8)`
```
harrisonmutai-arm: Instead of hard-coding the header size why not just refer to `te.hdr_size`?  It's slightly clearer what the intent is then.

source/transfer_list.rst
manish-pandey-arm: Yes, you are right. When the overwrite operation returns an abort, the caller may choose to fall back to the "Adding a new TE" procedure, at which point tl_base_addr will be explicitly provided as an argument to that call.

source/transfer_list.rst
manish-pandey-arm: We can do this universally when migrating to appendinx ?

jmarinho (dismissed):
LGTM

harrisonmutai-arm (approved):

jmarinho (commented):
LGTM

source/transfer_list.rst
```diff
@@ -417,6 +417,39 @@ Inputs:
    and coalescing consecutive `XFERLIST_VOID` entries into a single larger one. This can help
    reduce fragmentation and improve reuse of space in the Transfer List.
 
+Overwriting a TE
+^^^^^^^^^^^^^^^^
+
```
sjg20, line 426: It would be useful to define or motivate this more clearly. Why would one want to do this?

sjg20 (dismissed):
LGTM with one comment

jmarinho (approved):

source/transfer_list.rst
manish-pandey-arm: Overwriting a TE is a space-efficient way to update the data associated with an existing tag without changing its position in the transfer list. This is useful in scenarios where the tag ID remains valid, and only the payload needs to be refreshed. It avoids the overhead of allocating new space and modifying the TL structure unless the new data exceeds the original size, in which case a replacement strategy is triggered.

@sjg20 Do you these comments in spec or you happy with current version ?

manish-pandey-arm: @danh-arm, could you please merge this PR? @sjg20 is happy with the patch, aside from a single minor nit.

source/transfer_list.rst
```diff
@@ -507,6 +507,9 @@ Inputs:
 #. If `has_checksum`, add the sum of the 4 bytes from `new_tl_base + 0xc` to
    `new_tl_base + 0x10` to `tl.checksum` (`new_tl_base + 0x4`).
 
+.. note::
+   After relocating a TL, implementations should consider scrubbing the old TL memory if it contains
+   any secrets that might be accessible to later untrusted software.
```
danh-arm, line 556: minor: I think this note also applies in the case of Removing a TE or Overwriting a TE. This could perhaps be duplicated or placed in a more general section?

danh-arm (approved):
Just 1 minor comment that can be addressed later

PR #76, opened 2025-09-08
Update specification for transfer list with GPT info tag

xialin-arm:
There is a new entry to add in transfer list. This is for the primary GPT corruption logging, which use the transfer list to pass the information from BL2 to BL32.

### DIFF
--- a/source/transfer_list.rst
+++ b/source/transfer_list.rst
@@ -879,6 +879,9 @@ including TF-A, OP-TEE and Hafnium.
    * - :ref:`0x108 <tab_entry_point_info32>`
      - AArch32 Entry Point Info
 
+   * - :ref:`0x109 <tab_gpt_info>`
+     - GPT Error Info
+
 **OP-TEE pageable part address entry layout (XFERLIST_OPTEE_PAGEABLE_PART_ADDR)**
 
 This entry type holds the address of OP-TEE pageable part which is described in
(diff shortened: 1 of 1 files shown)


source/transfer_list.rst
```diff
@@ -1374,5 +1374,42 @@ subsequent images. It's usage is identical to the 64-bit form represented by
      - hdr_size + 0x20
      - Register R3.
 
+**Primary GPT corruption information (XFERLIST_GPT_INFO)**
+
+This entry holds an indication of whether the primary GPT is corrupted or not.
+The field `gpt_corrupted_info` is a 8 bits width field, there the bit [0] is
```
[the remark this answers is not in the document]
harrisonmutai-arm: I can't see this change?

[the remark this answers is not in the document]
xialin-arm: https://review.trustedfirmware.org/c/TF-A/trusted-firmware-a/+/40818/16/drivers/partition/partition.c

source/transfer_list.rst
```diff
@@ -1374,5 +1377,41 @@ subsequent images. It's usage is identical to the 64-bit form represented by
...
+
+   * - hdr_size
+     - 0x1
+     - 0x3
+     - |hdr_size_desc|
+
+   * - data_size
+     - 0x1
+     - 0x4
+     - The size of the GPT error information in bytes.
```
harrisonmutai-arm, line 1409: Considering the field is a fixed size, we should probably state here that this is always 1 in this version of the spec.

harrisonmutai-arm (approved):

source/transfer_list.rst
```diff
@@ -1374,5 +1377,41 @@ subsequent images. It's usage is identical to the 64-bit form represented by
      - hdr_size + 0x20
      - Register R3.
 
+**Primary GUID Partition Table (GPT) error information (XFERLIST_GPT_INFO)**
```
ManishVB-Arm, line 1380: Can this be XFERLIST_GPT_ERROR_INFO instead?

ManishVB-Arm, line 1380: Better remove primary so it can be scaled up later for any other error information?

source/transfer_list.rst
```diff
@@ -1374,5 +1377,41 @@ subsequent images. It's usage is identical to the 64-bit form represented by
...
+
+This entry holds an indication of the type of GPT error.
+The `gpt_error_info`` field is an 8-bit value. Bit 0 indicates corruption of
+the primary GPT. Bits [7:1] are reserved for future use.
```
ManishVB-Arm, line 1384: Primary GPT header to be specific.

harrisonmutai-arm (approved):

sjg20 (approved):

source/transfer_list.rst
```diff
@@ -1374,5 +1377,41 @@ subsequent images. It's usage is identical to the 64-bit form represented by
...
+
+   * - data_size
+     - 0x1
+     - 0x4
+     - The size of the GPT error information in bytes, it is always 1 byte.
+
+   * - gpt_error_info
+     - data_size
+     - hdr_size
+     - The `gpt_error_info` field contains GPT error information.
```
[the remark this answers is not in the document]
ManishVB-Arm: gpt_error_info has already been provided above at line 1383. The intention of this field is to capture all GPT-related error information, not just a single condition. At the moment, we are only targeting the primary GPT header error via Bit 0. We can certainly improve the log message to clarify that the primary GPT header is corrupted, so the secondary GPT header is being used.

Other bits in gpt_error_info are reserved for potential future use. If we later need to represent additional GPT error conditions, those can be encoded using the reserved bits.

PR #77, opened 2025-09-15
Add Trusted Boot Firmware Configuration entry

harrisonmutai-arm:
Defines a TE type for passing the Trusted Boot Firmware Configuration (TB_FW_CONFIG) DT blob. Includes layout table, usage description, and references the TF-A bindings. Note, this DT format is platform specific, unstable and only used by TF-A BL1/BL2. It's content is not expected to make it's way to subsequent firmware stages. A binding document is provided as reference for what contents might be included by a platform.

### DIFF
--- a/source/transfer_list.rst
+++ b/source/transfer_list.rst
@@ -883,6 +883,9 @@ including TF-A, OP-TEE and Hafnium.
    * - :ref:`0x109 <tab_gpt_info>`
      - GPT Error Info
 
+   * - :ref:`0x10a <tab_tb_fw_config>`
+     - DT Trusted Boot Firmware Configuration (``TB_FW_CONFIG``)
+
 **OP-TEE pageable part address entry layout (XFERLIST_OPTEE_PAGEABLE_PART_ADDR)**
 
 This entry type holds the address of OP-TEE pageable part which is described in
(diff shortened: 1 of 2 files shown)


source/transfer_list.rst
```diff
@@ -1374,5 +1377,43 @@ subsequent images. It's usage is identical to the 64-bit form represented by
...
+
+   * - data_size
+     - 0x4
+     - 0x4
+     - The size of the configuration file in bytes.
+
+   * - tb_fw_config
+     - data_size
+     - hdr_size
+     - Holds a Trusted Boot Firmware Configuration file in DT format.
```
sjg20, line 1415: Thanks for the link to the binding:

https://review.trustedfirmware.org/plugins/gitiles/TF-A/trusted-firmware-a/+/c6c882a42795da82b18d2d7dff1265bf2a1a6aab/docs/components/fconf/tb_fw_bindings.rst

I think this needs to be brought into this spec, since it doesn't follow the Linux devicetree schema.

Some comments on the schema:
1. As mentioned previously, the properties should have hyphens instead of underscores.
2. UUIDs should be in binary format, not string, right?
3. This is presumably ARM-specific so I think this should be clearly mentioned in the entry type (i.e. it would not be valid to use this on other archs)
4. Should load-address be 64-bit? You should really have #address-cells etc. to indicate this
5. It isn't obvious which node this properties are in. I suppose it doesn't matter, since there is a compatible string, so long as it isn't the root node
6. The compatible string should be <vendor,device> so better to use <vendor> than <plat>, which might be confusing
7. The binding should indicate what these things are for. For example 'Platform Secure Partition Content Certificate UUID' does not impart much information. What is the certificate for and how do you create it or use it?
8. For 'owner', could you enumerate value values? Is this just for logging or does it affect operation.
9. The beinding refers to a 'Chain of Trust Bindings' document. It isn't clear how we decide if this needed, but if it is needed, it should be described in this spec too.

As a general comment, rather than UUIDs, perhaps strings would be more helpful for humans?

sjg20 (changes requested):

source/transfer_list.rst
sjg20: Is there any update on this? It there a plan to upstream to Linux schema, or will it stay separate?

harrisonmutai-arm: Hi Simon, thanks for your comment. I’ve been considering how to incorporate some of your suggestions, but I haven’t had much time to work through them. One challenge is that there are external platforms that have been using this DT format for quite a while. When the DT was originally defined, it came with very few constraints - properties were added ad hoc, and it was intended solely as an IMPDEF mechanism for passing information between BL1 and BL2. Unfortunately, going back now to standardize it isn’t straightforward, what I've provided so far is really just a best-effort attempt to outline a schema.

harrisonmutai-arm: For now the plan is to keep this separate from the Linux schema, we don't envision a scenario where this DT leaves the bounds of TF-A.

sjg20: To deal with compatibility, you could change the compatible string. It would allow you to write new and old if you really need to (with new being the default).

Is there no interest in getting this through schema review?

jmarinho:
Hi @harrisonmutai-arm , should we move this PR to https://github.com/FirmwareHandoff/ProjectTEs ?

@manish-pandey-arm  @danh-arm

harrisonmutai-arm:
Hi @jmarinho, yes we should, moving the change over to the new project.

Issue #63, opened 2025-03-17
Spell out the rationale for requirement R4


danh-arm:
It's not obvious why R4 is important. This requirement along with the fact that TEs can only be added with the same or larger alignment, means that it's possible relocate the whole TL without each TE needing its own alignment setting and without needing shuffle the TEs around within the TL.

manish-pandey-arm:
PR merged to update the R4 with clearer wording

PR #83, opened 2025-11-25
Clarify TL header alignment field description

jmarinho:
Also clarify R4 that specifies TL relocation requairement around TL alignment.

Change-Id: I3a153377776b8f97a4733119c64bbd94a2237737


sjg20 (approved):

manish-pandey-arm (approved):

danh-arm:
FWIW this looks good to me. Are we happy this addresses https://github.com/FirmwareHandoff/firmware_handoff/issues/63? If so, we can close that.

PR #84, opened 2025-12-17
Version 2.0 release

jmarinho:
This commit updates the specification version to 2.0.
It also updates the version of the TL Header to 2.

### DIFF
--- a/README.md
+++ b/README.md
@@ -65,5 +65,5 @@ Status
 As of November 2025, version 1.0 is withdrawn and should not be used.
 Version 1.0 was withdrawn because products shipped with a TL header cheksum implementation that differed from the version 1.0 definition.
 
-A version 2.0 of the specification is currently under work and will be published shortly.
+Version 2.0 of the specification is the latest release.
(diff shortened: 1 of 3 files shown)


apalos (approved):

danh-arm (approved):
Looks good, but I thought we were going to resolve #66 (including moving the std ops to an appendix) before making the 2.0 release?

manish-pandey-arm:
#66 is ready, if it can be reviewed promptly ,  we can take this in release. Moving to appendix will be straight forward change. @harrisonmutai-arm has agreed to post one.

jmarinho:
> Looks good, but I thought we were going to resolve #66 (including moving the std ops to an appendix) before making the 2.0 release?

Yes, @manish-pandey-arm @danh-arm this PR is just putting things in place, but let's only merge it once PR66 is done and the transfer entries have been moved to the secondary spec.
Realistically, that will still take some time.
Let's get back to it in the new year.

sjg20 (approved):

